PR #267, opened 2014-12-16
PieChart highlight pop out math error for highlight pie slice

wogg:
The shape of the popup pie slice gets wonky when the size changes. Not sure what the goal there was using the sin / cos values to adjust size of arc, but it becomes an oval due to the math.

I have patched my fork to just equally enlarge the arc's box in every dimension so the shape of the original arc is preserved, rather than changed in ovoid manner.

This issue becomes especially apparent when pie slices exceed 90 degrees.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/PieChart.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/PieChart.java
@@ -239,12 +239,13 @@ public class PieChart extends PieRadarChartBase<PieData> {
                     continue;
 
                 float shift = set.getSelectionShift();
-                float xShift = shift * (float) Math.cos(shiftangle);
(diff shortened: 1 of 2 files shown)


wogg:
Can you let me know what the rationale here is?

As far as I can tell, the math you are using suffices to determine if the direction of shift is positive or negative (using the trig functions) but calculates the wrong sizes for the containing rectangle. All the should be done is make the containing rectangle equally larger in each direction keeping the same center point, which is what the patch does. Thanks for thinking about it.

PhilJay:
Your math is absolutely correct. I was unable to directly merge this pull request because it interfeared with some of the changes I recently made to the library!
However I included your code in my latest local commit and it will be online in the next release.

Thank you and regards,
Phil

PR #281, opened 2014-12-27
Functioning HorizontalBarChart

robertschmid-ast:
This is a Horizontal BarChart that extends from BarChart.  (Ideally, BarChart would be called ColumnChart).  There are several other minor changes that make MPAndroidChart more extendable and flexible.


R3curve:
Hey, there is a bug in HorizontalBarChart.java at line 90. mData.getXVals().get(0) is accessed without checking, if mData contains an element. You can experience the issue in HorizontalBarChartActivity, if you set the number of entries to 0.
I fixed the bug by replacing mData.getXVals().get(0) by "A".

Issue #262, opened 2014-12-15
java.lang.IllegalArgumentException: width and height must be > 0


shubhamb11:
Exception raised during rendering: width and height must be > 0
Exception details are logged in Window > Show View > Error Log

java.lang.IllegalArgumentException: width and height must be > 0
    at android.graphics.Bitmap.createBitmap(Bitmap.java:815)
    at android.graphics.Bitmap.createBitmap(Bitmap.java:794)
    at android.graphics.Bitmap.createBitmap(Bitmap.java:761)
    at com.github.mikephil.charting.charts.Chart.onSizeChanged(Chart.java:2197)
    at android.view.View.sizeChange(View.java:15703)
    at android.view.View.setFrame(View.java:15676)
    at android.view.View.layout(View.java:15592)
    at android.widget.RelativeLayout.onLayout(RelativeLayout.java:1076)
    at android.view.View.layout(View.java:15596)
    at android.view.ViewGroup.layout(ViewGroup.java:4966)
    at android.widget.FrameLayout.layoutChildren(FrameLayout.java:573)
    at android.widget.FrameLayout.onLayout(FrameLayout.java:508)
    at android.view.View.layout(View.java:15596)
    at android.view.ViewGroup.layout(ViewGroup.java:4966)
    at com.android.internal.widget.ActionBarOverlayLayout.onLayout(ActionBarOverlayLayout.java:494)
    at android.view.View.layout(View.java:15596)
    at android.view.ViewGroup.layout(ViewGroup.java:4966)
    at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1703)
    at android.widget.LinearLayout.layoutHorizontal(LinearLayout.java:1692)
    at android.widget.LinearLayout.onLayout(LinearLayout.java:1468)
    at android.view.View.layout(View.java:15596)
    at android.view.ViewGroup.layout(ViewGroup.java:4966)
    at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1703)
    at android.widget.LinearLayout.layoutVertical(LinearLayout.java:1557)
    at android.widget.LinearLayout.onLayout(LinearLayout.java:1466)
    at android.view.View.layout(View.java:15596)
    at android.view.ViewGroup.layout(ViewGroup.java:4966)
    at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1703)
    at android.widget.LinearLayout.layoutVertical(LinearLayout.java:1557)
[...]

PR #286, opened 2014-12-31
Fixes Issue #262 where it throws an exception when using the include xml tag

almilli:
This fixes an IllegalArgumentException (issue #262) when creating the draw bitmap with width/height of 0.  This is a valid case and also happens when using the include xml tag to include a layout that contains a chart.


JamesLee87:
Hey, How can I fix this problem? I have faced same problem...

JamesLee87:
Hi PhilJay, Could you upload fixed jar file for the version?

almilli:
@JamesLee87, you can git clone the master branch and build the jar yourself with maven.  Just run mvn package

PR #299, opened 2015-01-12
Update LineChart.java

Rajagopaul:
Assigning Alpha color for Cubic Fill was not available and now updated.


PhilJay:
Wow, thats a lot of performance overhead for fill-alpha!

PR #327, opened 2015-01-28
Added Android Studio gradle usage to Readme

Gopinathp:


FriesischScott:
I don't think this should be included as it is 55 commits behind the master.
I am fairly sure this guy did it: https://github.com/dodocat/MPAndroidChart

PR #336, opened 2015-02-03
Convert line endings from CRLF to LF.

johnjohndoe:


johnjohndoe:
@PhilJay Rebase onto master.

friederbluemle:
While I agree that files should have LF instead of CRLF endings, this **gradlew.bat** file is somewhat special as it is intended for the Windows operating system (which uses CRLF).
There even is an integration test in the Gradle project that ensures the correct platform specific line endings for the generated **gradlew.bat**: https://github.com/gradle/gradle/blob/master/subprojects/wrapper/src/integTest/groovy/org/gradle/integtests/WrapperGenerationIntegrationTest.groovy#L23

johnjohndoe:
This is not contrary to the fact that all files within the repository should be stored with Unix line endings aka `LF`. Git allows to set up the **configuration** in a manner that line endings are **converted to `CRLF`** on Windows machines on the fly when they are **checked out** from the repository. Storing files with `CRLF` or `LF` and `CRLF` is not recommend.

danielgindi:
Yes this file *should stay CRLF*. Phil I think this pull request can be closed...

johnjohndoe:
> ... This is a subtle but incredibly annoying fact of cross-platform work ... Git can handle this by auto-converting `CRLF` line endings into `LF` when you add a file to the index, and vice versa when it checks out code onto your filesystem ...

Please have a look at the [documentation](http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Formatting-and-Whitespace).
Keeping `CRLF` in files will discourage any non-Windows developer to contribute to your project. Tasks like checking out another branch or using *rebase* become awkward.

danielgindi:
Man if that's what stopping you, one *windows* file keeping its CRLFs, from
contributing, then really don't.
>
>
We know that git can handle crlfs, but we also know that in this particular
file it *shouldn't* be handled.

PR #345, opened 2015-02-09
Added android-maven plugin

jitpack:
With android-maven plugin MPAndroidChart can now be installed in the local maven repository:

    gradle install

In addition its now possible to get it as a maven dependency: https://jitpack.io/#jitpack/MPAndroidChart/v1.7.5 : )


jitpack:
Nice! With the next release that includes this patch it should be easy to download MPAndroidChart from https://jitpack.io/#PhilJay/MPAndroidChart/v1.7.5

vbauer:
:+1:

PR #373, opened 2015-02-24
Added gradle dependency information

jitpack:
So that MPAndroidChart can be easily added to gradle Android projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,15 +111,28 @@ Features
 Usage
 =======
 
-In order to use the library, there are 2 options:
+In order to use the library, there are 3 options:
 
 
 **1. jar file only**
  - Download the [**latest .jar file**](https://github.com/PhilJay/MPAndroidChart/releases) from the releases section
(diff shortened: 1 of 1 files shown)


PhilJay:
Thank you.

I have a short question:
Everytime I create a new release of this library, will it automatically be available in the public repository for everyone to use as a dependency? Or does that require any extra work except from updating the "Readme" file with the latest version?

jitpack:
Hi,
the nice thing is that no extra work is required. As soon as someone requests that new version JitPack will build it. Of course, if you like, you can make that first request to check that it works but generally if `gradle install` succeeds locally then it should on JitPack.

Maven information is correct. 

One thing to consider is a badge since it will show the latest release:
[Release](https://jitpack.io/#PhilJay/MPAndroidChart)

PhilJay:
Awesome, thank you!

PhilJay:
Sorry to bother you again, do you have any idea why the v2.0.1 causes build errors?

See here: https://jitpack.io/#PhilJay/MPAndroidChart

PR #422, opened 2015-03-09
Solve a null pointer exception

459522344:
Solve a null pointer exception


459522344:
Solve a null pointer exception

PhilJay:
Thank you, that has already been fixed on my local commit!

PR #429, opened 2015-03-10
CombinedChartRenderer - delegate renderer's creation to protected methods

Maragues:
As it is now, there's no way to use your own Renderers in a CombinedChart.

This PR simply delegates the Renderer creation to protected methods so that any class inheriting from CombinedChartRenderer can set custom renderers.


PhilJay:
The `CombinedChartRenderer` now has a method called `createRenderers(CombinedChart chart, ChartAnimator animator, ViewPortHandler viewPortHandler)` that allows the customized creation of renderers.

Maragues:
Great, thanks!

Issue #385, opened 2015-02-26
Accept List instead of forcing ArrayList


Maragues:
Constructors of *DataSet* (*LineDataSet*, *BarDataSet*) force you to pass an *ArrayList*. Is there any particular reason? I can create a PR replacing *ArrayList* for *List*.

PhilJay:
Well the reason is I like `ArrayList` nothing else :-D

Maragues:
@PhilJay you should always prefer interfaces over specific implementations, unless there's a good reason. They offer flexibility and decouple the code from a given class. There's a lot of literature about it

http://www.javaworld.com/article/2073649/core-java/why-extends-is-evil.html
http://uet.vnu.edu.vn/~chauttm/e-books/java/Effective.Java.2nd.Edition.May.2008.3000th.Release.pdf, item 18

My vote is to reopen the issue, but it's your code

PhilJay:
Allright, I'll do it as soon as there is the time. ´Currently it's not very high on my priority list tough :-)

PR #434, opened 2015-03-12
Use List instead of ArrayList as input parameter or return type

Maragues:
Solves issue https://github.com/PhilJay/MPAndroidChart/issues/385

I've gone through all the examples and they work ok, except for some that are also failing on your repo, which are

- CombinedChart - throws NPE, mRenderer is not initialized
- PieChart  - hole is draw black instead of transparent
- Realtime chart - throws NPE, mRenderer is not initialized
- Dynamic Data Adding - causes ANR

As I said, I've checked that these errors are also present in your master branch.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
+++ b/MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
@@ -9,6 +9,8 @@ import com.github.mikephil.charting.utils.Highlight;
 import com.github.mikephil.charting.utils.ValueFormatter;
 
 import java.util.ArrayList;
+import java.util.List;
+import java.util.List;
 
 /**
  * Class that holds all relevant data that represents the chart. That involves
@@ -45,10 +47,10 @@ public abstract class ChartData<T extends DataSet<? extends Entry>> {
     private float mXValAverageLength = 0;
 
     /** holds all x-values the chart represents */
-    protected ArrayList<String> mXVals;
+    protected List<String> mXVals;
 
     /** array that holds all DataSets the ChartData object represents */
-    protected ArrayList<T> mDataSets;
+    protected List<T> mDataSets;
 
     public ChartData() {
         mXVals = new ArrayList<String>();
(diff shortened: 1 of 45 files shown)


Maragues:
The NPEs in CombinedChart and RealTimeChart may be fixed by PR https://github.com/PhilJay/MPAndroidChart/pull/422

MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
```diff
@@ -95,41 +97,41 @@ public ChartData(ArrayList<String> xVals, ArrayList<T> sets) {
...
 
     /**
-     * Turns an array of strings into an arraylist of strings.
+     * Turns an array of strings into an List of strings.
      * 
      * @param array
      * @return
      */
-    private ArrayList<String> arrayToArrayList(String[] array) {
+    private List<String> arrayToList(String[] array) {
```
MGaetan89: `return Arrays.asList(arrays);`?

Maragues: Yeah, that'd make much more sense. I didn't go into analyzing improvements or if the code makes sense. Thanks!

PhilJay:
Thanks for this PR, unfortunately I cannot merge it because it interfears with a lot of changes I recently made.

Best regards,
Phil

friederbluemle:
This would still be nice to have, as it's generally better to use the interface of a collection type rather than a specific implementation. Maybe the PR can be rebased and reopened?

Maragues:
Thanks @friederbluemle , I was feeling quite alone :-) 

I can merge from upstream next monday and re-do the changes. Are you ok with that @PhilJay ?

Maragues:
@PhilJay I merged your master branch into mine, every example is working fine. Would you like to reopen this PR or should I create a new one? If you reopen it, I don't know if it'll include my latest commit.

PhilJay:
Thanks, just create a new one!
So now the lib takes list instead of arraylist as a parameter?

Maragues:
Yes, it'll accept any Object that implements List. In principle that doesn't change anything, since you store the index in every Entry instance, but the API is now more flexible and doesn't impose the ArrayList constraint.

PR #452, opened 2015-03-19
Arraylist to list

Maragues:
API classes take List parameters rather than ArrayList. All inner fields are declared List instead of ArrayList

See PR https://github.com/PhilJay/MPAndroidChart/pull/434, which was originally closed.


PhilJay:
Thank you, I just merged this on the CL to resolve some issues with my local commits!
Now `List` is supported instead of `ArrayList`.

Best regards,
Phil

PR #515, opened 2015-04-05
LimitLine support for XAxis

dexterbg:
+ centerViewTo method for BarLineChartBase


PhilJay:
Looks nice and clean, thank you! Did you test everything or should I have a look at it?

dexterbg:
I'm using it in a LineChart -- that should cover Bar & Scatter as well.
I haven't tested the horizontal bar chart version, but it's just a copy of the standard YAxis renderer.
A RadarChart implementation would not be that hard, but I doubt there's need for it.
Btw: I added the setTextStyle() method because FILL+STROKE text looks bad on dark backgrounds -- maybe worth a hint in the documentation.

dexterbg:
PS: there's an option to refactor the X/Y renderers into the AxisRenderer class, as the horizontal bar chart essentially only exchanges the data sources.

PR #524, opened 2015-04-07
Support for animation easing functions

danielgindi:
Demonstrated in PieChartActivity
Animations are now supported on all Android versions


PhilJay:
Thanks for taking the time! Looks awesome!

PR #547, opened 2015-04-13
Implement fling friction for the BarLineChartBase derived classes (Update 2)

mesropd:
Add an option to continue scrolling with friction after touch up (https://www.youtube.com/watch?v=9Z8UH8mB5Qg). 
In this update I create a MotionEvent and dispatch it via `dispatchTouchEvent` method, instead of calling perfomrDrag and getViewPortHandler().refresh() methods directly, so this is more adaptive to future updates.

Usage
```java
chart.setFlingFrictionEnabled(true);
chart.setFlingFrictionCoef(0.8f);
```

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -110,6 +110,19 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleData<? exte
 
     protected XAxisRenderer mXAxisRenderer;
 
+    /**
+     * If set to true, chart continues to scroll after touch up, but with friction, and its speed
+     * decrease over time, depend on mFlingFriction value
+     */
+    private boolean mFlingFrictionEnabled;
+
+    /**
+     * Fling friction coefficient in [0 ; 1] interval, higher values indicate that speed will
+     * decrease slowly, for example if it set to 0, it will stop immediately, if set to 1, it will
+     * scroll with constant speed, until the last point
+     */
+    private float mFlingFrictionCoef = 0.9f;
+
     // /** the approximator object used for data filtering */
     // private Approximator mApproximator;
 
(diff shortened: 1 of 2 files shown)


maximethebault:
It works just great! If I may suggest an improvement, maybe it would be better if fling fraction was disabled in case of pinch zoom, because we don't expect it to happen then. Other than that, well done!

mesropd:
You're right, I will fix it soon.

danielgindi:
@PhilJay I think this works :-)

danielgindi:
Although I think it should be enabled by default

danielgindi:
In the iOS version I'm implementing something similar, and chose the more generic name "dragDeccelarationEnabled"

danielgindi:
@PhilJay please let me know when you merge it - so I'll extend it to the pie/radar charts as well

PhilJay:
I would really love to merge it, but I have already tried it and it's unfortunately not very good concerning performance.

Also, in android there are other, more common ways of achieving a fling effect, but they require more effort to implement.

To be more specific, I am talking about [this example](https://android.googlesource.com/platform/development/+/master/samples/training/InteractiveChart/src/com/example/android/interactivechart/InteractiveLineGraphView.java) from Google.

Please note the used `OverScroller` class in that example.

danielgindi:
Well I haven't noticed any performance penalty but I'm on a simulator....
Notice that they are not doing anything different actually... they are using the fling gesture recognizer to fetch the velocity (on the inside the calculation is the same), and when done panning it fires off a mission that moves the view as necessary using the same math.

The difference, and I think what gave poor performance for custom animation implementaions also- is that on the devices you need to sync with the recommended FPS of the specific device. This is done by the Choreographer class which is used by Animation classes and other classes, available only since API 11 or 12 I think.

So we can choose to implement the standard gesture recognizers, or leave it as it is and use the Choreographer to control the frame rate (or Google's way, which essentially does the same thing).

That's what I did on iOS - using the standard UIGestureRecognizers, except for radial charts because the rotation is offset by the the legend etc. so we cannot use the angle to the view's center.

Btw, I rwally hate Android's blue stuff at the end of scrolling, but maybe it's because I'm coming from iOS... I love the bouncing, and I think I'll try to implement it in the line-based charts. Maybe we need to keep this detIl platform specific- OverScroll for android and bounce for iOS?

danielgindi:
Dammit, OverScroller is only supported in API >= 9

danielgindi:
I have an implementation ready, here: https://github.com/PhilJay/MPAndroidChart/pull/612

@mesropd It's kind of an evolution of what you did, only going through the changes I've done in iOS version then the adjustments for Android methods according to Google recommendations, so you might not recognize it ;-)

mesropd:
I see the new implementation of deceleration ant its more interesting that mine, thanks :)

Issue #348, opened 2015-02-11
Release apk - Proguard - No animations?


raulfdezhdez:
Hi!
First of all, thank you for your work. Excellent library.

I have one little problem when exporting a signed apk with proguard. When opening chart views, animations are not happening.
If I turn off proguard, the animations start to work again.

I am testing in two Android 5.0.1 phones (Nexus 6 and Samsung S5)

Any idea?

Thanks in advance.

PhilJay:
Pffff. I honestly do not have the slightest idea why this is (obviously) not working.
Does anyone else know something?

Bishalsingh:
The trick is the fact that the methods called by the PropertyValuesHolder are either set* or get*, so the ProGuard rules to keep such methods are:

void set*(***);
*** get*();
So, based on that, if your animated target is only one of a few classes, add this per animated target class:

-keep class my.package.MyClass {
    void set*(***);
    *** get*();
}
If all your targets extend some base class, you can add a single rule to save them all:

-keep class * extends my.package.MyBaseClass {
    void set*(***);
    *** get*();
}

Bishalsingh:
Can I see your Logcat @raulfdezhdez ??
In mine case the problem was due to incompatibility issues and i fixed it as:
 ActionBarSherlock was failing to find a constructor "a(Activity,int)" using reflection.
 ProGuard doesn't (can't) know this and has removed the constructor, because it appears unused in the code. You therefore have to explicitly tell ProGuard to keep it:

-keepclassmembers class * extends com.actionbarsherlock.ActionBarSherlock {
    <init>(android.app.Activity, int);
}

raulfdezhdez:
Mine is not generating any error log signing the apk or at runtime, and I am not using SherlockBar.
I have tried with the typical proguard rules, but no luck.
At the moment it is not a big problem and I am not in a hurry, so I will keep trying and if I find any clue o solution I will post it.
Thanks for the great library!

MarcSalvat:
I had the same problem. I excluded the library from proguard adding this to proguard rules:

-dontwarn com.github.mikephil.**
-keep public class com.github.mikephil.** {
  public protected *;
}

I'm not sure if there's a better approach but this works

romangromov:
@MarcSalvat this way MPAndroidChart become not obfuscated by proguard

mesropd:
Here is the reason. Original ChartAnimator.java contains this code
```java
[...]

PR #590, opened 2015-04-25
Readme: add notice about Proguard

Evisceration:
* When using proguard and NOT whitelisting MPAndroidChart, some functionality
    will break like animate methods.

Change-Id: If329842301eb41f7f713c1148b5ca364917e3d7a
Signed-off-by: Alexander Martinz <eviscerationls@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,15 @@ dependencies {
  - Import the library folder (`MPChartLib`) into your Eclipse workspace
  - Add it as a reference to your project: [referencing library projects in Eclipse](http://developer.android.com/tools/projects/projects-eclipse.html#ReferencingLibraryProject)
 
+Proguard
+=======
+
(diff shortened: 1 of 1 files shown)


Evisceration:
This was something old i had lying around in my local branches.

I just noticed that you have updated the animation code since i had to whitelist it at proguard.

Will recheck if this still applies :)

Evisceration:
Yes, unfortunately it still applies.

W/PropertyValuesHolder﹕ Method setPhaseX() with type float not found on target class class com.github.mikephil.charting.animation.ChartAnimator

just whitelisting the ChartAnimator works as well.
-keep class com.github.mikephil.charting.animation.ChartAnimator { *; }

mesropd:
Right fix, and here is the explanation why it breaks )
https://github.com/PhilJay/MPAndroidChart/issues/348

Evisceration:
alright, as i see this was noticed before already (sorry for not going through all the closed issues :))

PR #608, opened 2015-04-30
Bubble chart implementation, ported from the iOS version :-)

danielgindi:


PhilJay:
Wow, this looks fantastic, I'm already eager to try this out! Thank you! :-)

danielgindi:
:-)

danielgindi:
You now have the task of updating the docs and screenshots... Lol

PhilJay:
Yep, already on it :-)
I'll post the screenshots as soon as it's in a release version, so probably sometime next week.

I just tried them out and they look cool. I have a question concerning the size of the circles: How exactly is the actual size when drawn determined?

danielgindi:
Well there's a range of size values that the user specifies, and the largest size must fit in its column without overlapping with those beside it.
So (chartWidth / xValueCount) * (entrySize / maxSize).

If you have 500px wide chart, and only two values on the x-axis,
Then 250px is the max size for the bubble. Then you apply the relative size of the individual entry...
Quite simple actually!

PhilJay:
I have discovered a minor issue with the bubble sizes.
It seems the sizes are calculated correctly when the chart is in portrait mode (height > width) but when the chart is in landscape mode (width > height) the bubble sizes are too small.

You can observe this in the example of the bubble chart.

danielgindi:
Well it does try to take both height and width into consideration, which I
supposed is wrong? We could easily remove the Math.max there, but we need
to be sure. I'll look at some images of such charts and then I'll know
more...

danielgindi:
By the way, the Square Root there - is a standard for compensating for the way the brain interprets the bubble sizes. All Bubble Charts are supposed to be calculated this way so the brain can detect the correct difference between bubbles by their sizes, and the docs should say that the bubbles are `sqrt`ed...

PhilJay:
What do you think of this approach?

```java

protected void drawDataSet(Canvas c, BubbleDataSet dataSet) {

        Transformer trans = mChart.getTransformer(dataSet.getAxisDependency());

        float phaseX = mAnimator.getPhaseX();
        float phaseY = mAnimator.getPhaseY();

        List<BubbleEntry> entries = dataSet.getYVals();

        Entry entryFrom = dataSet.getEntryForXIndex(mMinX);
        Entry entryTo = dataSet.getEntryForXIndex(mMaxX);

        int minx = Math.max(dataSet.getEntryPosition(entryFrom), 0);
        int maxx = Math.min(dataSet.getEntryPosition(entryTo) + 1, entries.size());
        
        widthBuffer[0] = 0f;
        widthBuffer[2] = 1f;
        trans.pointValuesToPixel(widthBuffer);
        
        // calcualte the full width of 1 step on the x-axis
        final float maxBubbleWidth = widthBuffer[2] - widthBuffer[0];

        for (int j = minx; j < maxx; j++) {

            final BubbleEntry entry = entries.get(j);

            pointBuffer[0] = (float) (entry.getXIndex() - minx) * phaseX + (float) minx;
            pointBuffer[1] = (float) (entry.getVal()) * phaseY;
            trans.pointValuesToPixel(pointBuffer); 

            final float maxSize = (dataSet.getMaxSize() != 0f) ? dataSet.getMaxSize() : 1.0f;
            final float factor = entry.getSize() / maxSize;
            final float shapeHalf = (maxBubbleWidth * factor) / 2f; 

            if (!mViewPortHandler.isInBoundsY(pointBuffer[1]))
                continue;

            final int color = dataSet.getColor(entry.getXIndex());

            mRenderPaint.setColor(color);
            c.drawCircle(pointBuffer[0], pointBuffer[1], shapeHalf, mRenderPaint);
        }
    }
```

What this basically does is calculate the maximum width the bubble can have on the x-axis (1 step), and then scale all smaller bubbles down.

One other thing I have noticed is that the maximum size value is only `DataSet` specific, and not for the whole data object. Which means that a bubble-size of 100 in one dataset could be equally sized to a bubble size of 10 in another dataset, assuming that they are the maximum values in each dataset. Is that correct?

danielgindi:
It is correct- different DataSets usually contain different kind on information, so it's not a problem...

Your implementation seem to be doing the same thing it did previously, only on the x axis only...
I'm not sure yet that we should ignore the height, as the biggest bubble still should't overflow on the y axis,
But the bigger problem here is that you did not `sqrt` the size...

PhilJay:
Well the overflow thing is also an issue yes. Why exaclty does the size have to be sqrt`ed?

danielgindi:
The human visual system
<http://en.m.wikipedia.org/wiki/Human_visual_system> naturally
experiences a disk's size in terms of its area
<http://en.m.wikipedia.org/wiki/Area>. And the area of a disk—unlike its
diameter <http://en.m.wikipedia.org/wiki/Diameter> or circumference
<http://en.m.wikipedia.org/wiki/Circumference>—is not proportional to its
radius <http://en.m.wikipedia.org/wiki/Radius>, but to the square of the
radius. So if one chooses to scale the disks' radii to the third data
values directly, then the apparent size differences among the disks will be
non-linear and misleading. To get a properly weighted scale, one must scale
each disk's radius to the square root
<http://en.m.wikipedia.org/wiki/Square_root> of the corresponding data
value *v*3 This scaling issue can lead to extreme misinterpretations,
especially where the range of the data has a large spread. And because many
people are unfamiliar with—or do not stop to consider—the issue and its
impact on perception, those who are aware of it often have to hesitate in
interpreting a bubble chart because they cannot assume that the scaling
correction was indeed made. So it is important that bubble charts not only
be scaled in this way, but also be clearly labeled to document that it is
area, rather than radius or diameter, that conveys the data.

http://en.wikipedia.org/wiki/Bubble_chart

PhilJay:
Okay, I have made some adjustments and it now takes width and height into consideration but fully scaling up even in landscape mode.

I think we are ready for a new release, what do you say?

danielgindi:
Did you SQRT it?
Yeah I think we are there just let me a few hours to look at it :-)

PhilJay:
Yep I did (or I think I did), I am not that familiar with Bubble-charts so wont do any harm if you take a second look at it. :-)

PhilJay:
By the way: Have you recently checked the `ios-charts` tag I created on stackoverflow:
http://stackoverflow.com/questions/tagged/ios-charts

It's already getting a lot of attention.

danielgindi:
Okay we have a problem... When zooming in the bubbles do not keep their relative sizes

danielgindi:
When zooming on X only

PhilJay:
You mean that the bubbles get bigger?
Upon zooming?

danielgindi:
They should get bigger - but when zooming on X only - they get too big in relation to the others

PhilJay:
They always occupy a maximum of exactly 1 step on the x-axis.

danielgindi:
Yeah but I'm not sure that's right. Maybe add an option to choose to have it relative to Y axis too?

PhilJay:
As soon as the y-axis becomes the limiting factor, it takes over.

danielgindi:
It actually is never limiting it on the Y axis!
Because you are taking min-y and max-y which will always be the edges of the Chart, even beyond the viewport, when zoomed in...

PhilJay:
Well I see what you mean. How would you change it?
But when I zoom the x-axis only, it limits up to a point, and then the y-axis limits.

danielgindi:
It limits up to the x axis point - I don't see any y axis limit...

PhilJay:
You can observe it when the bubbles are suddenly significantly smaller than 1 step on the x-axis.

PhilJay:
But thats only in landscape mode i see.

danielgindi:
Yeah, that's because in Landscape the Y axis is short :-)

danielgindi:
How can we calculate the actual distance between two Y-lines - without hacking access to the YAxis renderer computed data?

PhilJay:
Well with the transformer you can transform the bounds of the viewport (top and bottom) to y-values, and then derive the currently visible range from them

PhilJay:
It doesn't contain information, but it can calculate the y-position the top and bottom edge of the chart would have. You understand what I mean?

danielgindi:
Yes but I'm thinking bout limiting the heigh to one y-line height.
Also in some cases it could be too small
Can you think of another solution?

נשלח מה-iPad שלי

‫ב-4 במאי 2015, בשעה 23:59, ‏‏Philipp Jahoda ‏<notifications@github.com> כתב/ה:‬

PhilJay:
Please try what I just pushed. I'm not sure but maybe this is what you meant :-)

danielgindi:
Well, not perfect but reasonable :-)
I think the best algorithm to choose the limits will eventually be setter properties for the user to adjust...

Why did you revert the Transformer changes? The count base on phasing is not correct for the bubble chart as the animation is different - we always draw all the label on the viewport between minx and maxx - but changing their positions based on phaseX/Y

PhilJay:
I resetted them because it immediately crashed when I started the bubble demo 😊

danielgindi:
Weird. Didn't crash for me... But it should be to-from, without phase...
Otherwise the labels are misbehaving in animation!

PhilJay:
Weird, Ill test it again 👌

PhilJay:
You coded `from-to`, isn't it `to-from`?

danielgindi:
Oh yeah!
Btw what about implementing the approximation filters back in?

‏בתאריך יום שלישי, 5 במאי 2015, Philipp Jahoda <notifications@github.com>
כתב:

PhilJay:
Good idea. I started this about a year ago, but was too busy doing other things so I never had the change to finish it. But would definitely be a great idea.

danielgindi:
Yeah it will spare a lot of coding every time for approximations...
I think we can release a version now right?

‏בתאריך יום שלישי, 5 במאי 2015, Philipp Jahoda <notifications@github.com>
כתב:

PhilJay:
Okay new release is out :-)

Next things on my list would be the following (big) topics:

 - getting away from x-axis indices (instead double)
 - y-values as double
 - approximation

danielgindi:
Yay! I'll release too :-)

PR #612, opened 2015-05-01
Deceleration after dragging in Line and Pie/Radar charts

danielgindi:
You should test this for performance!

Anyway I've looked at the `OverScroll` (API>=16) implementation, and Google says that to work with it you need to `postInvalidateOnAnimation()` and listen to `computeScroll()`.

The `OverScroll` is API 16 and is a bit of an overkill and meant for views, so I haven't used it, but used the same mechanism (`postInvalidateOnAnimation`/`computeScroll`).
If we will want to support bounce on edges, we will need to support it in the viewport and maybe in the renderers. `OverScroll` is meant to do that on `View`s only...

Also I've used `VelocityTracker` (which is used by the Fling recognizer) because we are not tracking fling but tracking sequential movement for panning. So android supplies the velocity arguments in the wrong listener but at least the `VelocityTracker` is public...

I can't test to see how smooth it works because the simulator only generates 10 fps when using `postInvalidateOnAnimation` which is incorrect, and I don't have a device here.

### DIFF
--- a/MPChartExample/src/com/xxmassdeveloper/mpchartexample/notimportant/MainActivity.java
+++ b/MPChartExample/src/com/xxmassdeveloper/mpchartexample/notimportant/MainActivity.java
@@ -59,7 +59,7 @@ public class MainActivity extends Activity implements OnItemClickListener {
         setContentView(R.layout.activity_main);
 
         // initialize the utilities
-        Utils.init(getResources());
+        Utils.init(this);
 
         ArrayList<ContentItem> objects = new ArrayList<ContentItem>();
 
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -95,6 +95,16 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleData<? exte
 
     protected boolean mDrawBorders = false;
 
+    /** If set to true, chart continues to scroll after touch up */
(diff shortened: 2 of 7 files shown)


PhilJay:
This is what I expected! Nothing more to say than GREAT JOB!

PhilJay:
Maybe there is some improvement potential for the PieChart and RadarChart in terms of expected behaviour. It seems like sometimes the deceleration triggers into the opposite direction of the fling gesture.

But other than that, I could not see any issues even on lower android versions. :+1:

danielgindi:
Yeah it's kind of trickey- because the angle does not have a direction...
Calling setRotation- will move clockwise or counter?

‏בתאריך יום שישי, 1 במאי 2015, Philipp Jahoda <notifications@github.com>
כתב:

PhilJay:
Yes, I saw your comments in code about that.

If the "mRotationAngle" property is increased, the chart will move clockwise.

danielgindi:
Yes I'm working on it. It's not that simple but I think I have a solution

danielgindi:
Ok, I've come to a conclusion that there's no shortcuts in this...
The only reason that we know with our eyes what is the direction that the circle moved, is because we've seen it move.
When we watch a wheel go around, the brain tracks the rotation and as long as it can catch a movement of less than 180 degrees per frame - it assumes that it moved in that direction. If it's more than 180 - the brain/eyes get confused.

So what I'm doing is building a velocity tracker that tracks the *angles* instead of the *points*, and while tracking it will look at the direction. The built-in VelocityTracker can't do that as it tracks points.

I'll post the changes soon!

PR #614, opened 2015-05-01
Large value formatter now can appends a specified text to the result string

aint:
Now, for adding a dollar-sign to yAxis (for example) we need only use a non default constructor of ```LargeValueFormatter```:
```java
yAxis.setValueFormatter(new LargeValueFormatter("$"));
```


MPChartLib/src/com/github/mikephil/charting/utils/LargeValueFormatter.java
```diff
@@ -11,25 +11,34 @@
...
+     */
+    public LargeValueFormatter(String text) {
+        this();
+        mText = text;
+    }
+
     @Override
     public String getFormattedValue(float value) {
-        return makePretty(value);
+        return makePretty(value) + " " + mText;
```
MGaetan89: If `mText` is `null` or empty, this would add an trailing space at the end, no?

PR #694, opened 2015-05-26
Auto scale y axis

danielgindi:


PhilJay:
Can you in short explain what that feature does?

danielgindi:
Yes :-)
When you change the viewport to show only a subset of the data (for example by scaling on the x axis) - it causes the y axis to scale to that subset of data.

You can see an explanation here: https://github.com/danielgindi/ios-charts/issues/89

danielgindi:
This feature was proposed by @AlBirdie which works with financial charts, and is required by him. So this opens up more possibilities for those working with financial charts.

Bhupendra-tbox:
Hi Daniel, Can we get max & min value after change autoScaleMinMaxEnabled feature?

jlo1:
@danielgindi  curious if you have any thoughts/fixes for this issue here: https://github.com/PhilJay/MPAndroidChart/issues/3374

Thanks!

PR #695, opened 2015-05-27
Allow controlling the extra offsets caused by the legends

danielgindi:


PhilJay:
I like the feature, but maybe we can include this more into the `ViewPortHandler`?

danielgindi:
Well the viewport handler has offsets, but the problem is that the chart is auto-calculating offsets based on legends - and you might want to modify that relative to whatever calculation is done.

And because legends are not the only offsets applied, and there are also minimum offsets etc. - you will want in most cases to just append to the offsets and not replace them in the viewporthandler...

danielgindi:
Do you really not see the point in this? I have people (and I) asking fir an option to adjust the legend's spacing- while keeping the automated calculation...
When do not really want to calculate the size of the legend yourself, but you decide that the spacing is too much or not enough

PhilJay:
I see it, im just evaluating :-)
I will merge it asap :-)

PhilJay:
How does this interact with the already existing `setViewPortOffsets(...)` method?

danielgindi:
It is passed through `restrainViewPort` in `calculateOffsets`.
`setViewPortOffsets` is an alternative, that sets your custom offsets - without accounting for legends etc. - that's for when you really know what you are doing.
But in most cases, you want the chart to calculate the offsets itself, and then add some

PhilJay:
Okay, I see. Makes sense.

PR #707, opened 2015-05-30
Flag to prevent overlapping y-values in line charts

oliver-jonas:
With draw values enabled, depending on the horizontal zoom setting and the actual data in the data-set, value labels may overlap (over-draw each other), meaning it's hard to distinguish values or you cannot make out any values at all.

I have added a drawValuesOverlap flag, which defaults to true (no change), which by setting to false prevents overdrawing. This takes into account both horizontal and vertical size and positioning of value labels.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -402,9 +402,9 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleData<? exte
                         .getPaintAxisLabels());
             }
 
-            float xlabelheight = mXAxis.mLabelHeight * 2f;
+            if (mXAxis.isEnabled() && mXAxis.isDrawLabelsEnabled()) {
 
-            if (mXAxis.isEnabled()) {
+                float xlabelheight = mXAxis.mLabelHeight * 2f;
 
                 // offsets for x-labels
                 if (mXAxis.getPosition() == XAxisPosition.BOTTOM) {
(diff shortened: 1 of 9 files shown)


Ras-Tafari:
Would this work with stacked bar chart also? As I have a problem when I have `barChart.setDrawValuesForWholeStack(true)` enabled and one of the stacks which precedes another one is zero, the values overlap. Example: i have 3 stacks and two values on X axis. when the bar entries looks like: new `BarEntry(new float[]{1, 2, 3}, 0)` and the second `new BarEntry(new float[]{2, 0, 1}, 1)`, the first case will be drawn properly, but in the second one i'll get 0 overlapping with 1. I think it would be nice to just supress drawing zeros.

oliver-jonas:
When using the cubic line chart with a data-set that is not uniformly distributed along the x-axis (e.g. when using it for a time-based chart and no data was available at a given time interval) the cubic lines may go backwards or even loop.

It's easy to reproduce this with CubicLineChartActivity by not adding a value to the data-set when i > 5 and i < 25. See image: http://i.imgur.com/9yB4OBh.png

I've committed a change to limit the x-intensity on prevDx and curDx relative to the given intensity. See image: http://i.imgur.com/YjOk19z.png

oliver-jonas:
@Ras-Tafari: Not overdrawing labels for the stacked bar charts would need to be handled in the BarChartRenderer. I've only updated the LineChartRenderer.

Ras-Tafari:
Okay, I'll try to dig something on my own then, thanks for the advice about where I should start. Basically it should be easy, I'll just put some `if(...)` statement to check if the sum is zero while drawing the values for stacked bars, as I don't think so there is a need to display them in stacked bars.

oliver-jonas:
Hi Philipp, have you looked at these two commits? Do you need anything to resolve the merge conflict (I'm not sure why the previous commits are showing). Regards.

PR #708, opened 2015-06-01
New feature - suppress overlapping values

Ras-Tafari:
I have encountered a problem when some of the stacks sums in stacked bar chart were zero, they were overlapping with other stacks values. So I have made some changes to allow the user suppress drawing zero values (since the stack won't be visible and that should be enough for the user there was no data). To do this all the user has to do is to use `setDrawSuppressZeroSums(boolean enabled)` on either `BarChart` or `CombinedChart`.

PS. Sorry if I have left any mess around, I have never been using GitHub : (

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarChart.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarChart.java
@@ -38,6 +38,12 @@ public class BarChart extends BarLineChartBase<BarData> implements BarDataProvid
      * just their sum
      */
     private boolean mDrawValuesForWholeStack = true;
+    
+    /**
+     * if set to true, sums, which are zero are not drawn on stacked bars.
+     * It is used to prevent overlapping values on the chart.
+     */
+    private boolean mDrawSuppressZeroSums = false;
 
     /**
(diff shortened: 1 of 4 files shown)


PhilJay:
Can't this be done with the valueformatter:

```java
if(value <= 0f) return "";
```

Ras-Tafari:
Tried this and it also works : ) Well didn't know it can be achieved by this as well,. Anyway i think drawing zero should be suppressed by default as it makes no sense to display it on the stacked bar when other stacks have values. So it is now up to you whether you want to keep my changes or not. Cheers!

danielgindi:
There are other cases where it is not ideal - think of a "1" value - the bar will still be too small to show the label. So just hiding zeros is not the proper solution. We may want to do it conditionally on whether there is space for drawing the values or not.

Ras-Tafari:
Well true, it also depends on the graph, in some cases (like mine) where '1' is quite a big value, just hiding zeros helps. And probably there should be other solution which as you said should take into account the space for drawing and handle it somehow can't think for any ultimate solution right now. Ok, since the situation can be more complex I have renamed the title.

PR #710, opened 2015-06-01
Improvements to legend drawing

danielgindi:
Line spacing is used correctly now.

Previously there was use of YEntrySpace, but now we are taking the correct line spacing, and defaulting YEntrySpace to 0.

To solve the legend form alignment issue to the text line - I still used a text height calculation, as some fonts (namely the Open Sans) do not have the "A" stretched from ascent to descent, but do have other letters ("J", "Q", "`") going out into the ascent or descent area. So calculating size of ABC instead of AJQ seems to be better in that matter. Taking half an A back from the baseline - is a good middle point for the form center.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/components/Legend.java
+++ b/MPChartLib/src/com/github/mikephil/charting/components/Legend.java
@@ -82,7 +82,7 @@ public class Legend extends ComponentBase {
     /**
      * the space between the legend entries on a vertical axis, default 5f
      */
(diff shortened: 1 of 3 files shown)


PhilJay:
Thanks, looks good, I'll test it! :-)

PhilJay:
Just tested it and works like a charm, very good.
There are some minor issues with the PieChart causing the legend to overlap with the chart if placed on the right, but this was an issue that has existed before.

PR #714, opened 2015-06-02
Allow to `setHighlightEnabled` on a single DataSet

danielgindi:
https://github.com/danielgindi/ios-charts/issues/114

By the way - highlightIndicatorEnabled had absolutely no meaning - but a duplicate of highlightEnabled...


PhilJay:
So highlighting can now be set via dataset?
And there is still a "global" highlight flag for the chart? I think we can also remove that one then :-)

PhilJay:
Ah I see it's gone as well!

danielgindi:
Yeah :-)

PR #737, opened 2015-06-10
Add option to invert fill area on LineChart

joshfriend:
Turns out looking like this:

example

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -1494,33 +1494,35 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleData<? exte
         public float getFillLinePosition(LineDataSet dataSet, LineData data,
                 float chartMaxY, float chartMinY) {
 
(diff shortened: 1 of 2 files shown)


PhilJay:
Thank you for your pull request, but this is a functionality that I do not want to handle via dataset.
If someone wants to invert the filled line, he can do so by creating a custom implementation of the FillFormatter.

If you would like to have this feature in the library, I will of course accept a pull request that includes an implementation of the FillFormatter that inverts the fill line, called e.g. `InvertedFillFormatter`

joshfriend:
Thanks for the feedback. I thought of that method but did not go that route because individual instances of `LineDataSet` may already set their fill properties (color, alpha, etc.) separately from each other. Therefore it made sense (to me) that it should also be possible to enable inverted fill per data set. 

If this is still unacceptable to you, I can add an `InvertedFillFormatter` class. Or, if you have alternate suggestions for a per dataset fill inversion setting, I'd be happy to pursue that as well.

PhilJay:
a9cb766310048db828a5d741095e2911c0f9809f

Now possible.
Wiki: https://github.com/PhilJay/MPAndroidChart/wiki/FillFormatter

PR #744, opened 2015-06-11
Highlight bugfixes when y values are possibly omitted

danielgindi:


PhilJay:
What exactly does that do? :-)

danielgindi:
If you omit some values in the yVals array (Take for example the BarChart sample, and let the data loop start from 10 instead of 0), it all works fine, except Highlights.
What happens is `getEntryForXIndex` returns the *closest* entry to the selected xIndex, which would result in a Highlight drawing when tapping a place where there's no value. And that highlight will take the height of the closest bar...

And then there's an issue with some highlights refusing to draw, depending on the xIndex. That's because of the `< yValCount` rule - where the yValCount does not actually represent any relation to the x indexes

PhilJay:
Allright, got it, nice fix!

PR #798, opened 2015-06-24
Support for negative values in stacked vertical bar chart

danielgindi:


danielgindi:
Re-opened the PR for re-merge after the re-base we had to do...

PR #800, opened 2015-06-24
Removed duplicate code

danielgindi:
Testing for `mInverted` in the main `if` or inner `if` do not have a meaning performance-wise, as in both cases it's just one boolean check.


danielgindi:
I know we've talked about it in the past and you don't have to merge it - but I saw that you stated you chose performance over readability. And I did something different this time, I used a single `if` instead of inline `mInverted ? ... : ...`, so there's the exact same performance now ;-)

danielgindi:
It becomes crucial to remove duplicate code when we're actually changing the code there and we have to be careful to attend to the duplicates too

PhilJay:
I see your point, it just looked so good before :-)

PR #1029, opened 2015-09-14
Stop using iterator to reduce memory allocation.

xizzhu:


danielgindi:
Did you measure what difference does that actually give?

xizzhu:
@danielgindi Depends on the data size, it's slightly better.

danielgindi:
It's micro optimization and it means nothing

‏בתאריך יום שישי, 2 באוקטובר 2015, Steven Zhu <notifications@github.com>
כתב:

PR #1058, opened 2015-10-01
Make skipWebLineCount consistent

danielgindi:


MPChartLib/src/com/github/mikephil/charting/charts/RadarChart.java
```diff
@@ -358,10 +358,7 @@ public void setDrawWeb(boolean enabled) {
      */
     public void setSkipWebLineCount(int count) {
 
-        if (count < 0)
-            count = 0;
-
-        mWebModulus = count + 1;
+        mSkipWebLineCount = Math.max(0, count);
```
PhilJay: Can't that result in 0?
if someone sets the skipWebLineCount to 0 - then the loop that skips weblines has += 0

danielgindi: I changed it to add 1 before the loop

The user says "skip 1 line", and you take that value and add 1 in the
setter, and subtract 1 in the getter - which makes nosense. Why not jusr
store the value...

‏בתאריך יום חמישי, 1 באוקטובר 2015, Philipp Jahoda <notifications@github.com>
כתב:

PhilJay: Looks good now!
Thanks!

PR #1064, opened 2015-10-02
Improved LineChart non-cubic fill rendering

tinsukE:
According to Romain Guy (http://stackoverflow.com/a/15208783/477243), Path drawing is not hardware accelerated, but drawn to a Bitmap and pushed to the GPU as a texture. This, on certain scenarios, causes the "W/OpenGLRenderer: Path too large to be rendered into a texture" error.

The proposed solution is to simply set the fill Path as a clip and then fill the clipped Canvas with the proper color.

According to http://developer.android.com/guide/topics/graphics/hardware-accel.html, clipping with a Path is hardware accelerated and should cause no performance penalties.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/renderer/LineChartRenderer.java
+++ b/MPChartLib/src/com/github/mikephil/charting/renderer/LineChartRenderer.java
@@ -347,22 +347,19 @@ public class LineChartRenderer extends LineScatterCandleRadarRenderer {
             int maxx,
             Transformer trans) {
 
-        mRenderPaint.setStyle(Paint.Style.FILL);
-
-        mRenderPaint.setColor(dataSet.getFillColor());
-        // filled is drawn with less alpha
-        mRenderPaint.setAlpha(dataSet.getFillAlpha());
-
         Path filled = generateFilledPath(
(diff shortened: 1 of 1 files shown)


PhilJay:
Looks good, I did some other adjustments locally already so I unfortunately cannot merge this.
I also implemented it for cubic-filled-paths, or did you see any issues with that?

tinsukE:
Since the cubic paths are being drawn into mDrawBitmap, a memory bitmap (not into a texture to be uploaded to the GPU), it is not needed there, although that could speed up the drawing if the operation is hardware accelerated.

I've also noticed that on LineChartRenderer.drawData(), mDrawBitmap is always allocated, when it is only needed if any of the DataSets is cubic. For purely linear charts, it would be a waste of memory. Would you like me to send a pull request to patch that too?

PhilJay:
Yes that would be great if you can implement it in a clean way and test it proberly! Please wait unitl I push for the next time (should be in a couple of minutes) before starting to implement this. Then you will save yourself some time merging. Thank you

PhilJay:
Alright, I just push you can go ahead now.

PR #1065, opened 2015-10-02
Make the ChartXAxisRenderer more flexible: now possible to overwrite drawing the line or label of the ChartLimitLine

forkch:
Hi Philipp

I have done a few refactorings of the ChartXAxisRenderer that allow the following:

It is now possible to extend the renderer and overwrite selectively the label or line drawing method for a ChartLimitLine (in my project I needed e.g. to draw a triangle as arrow at the top in one case or a short line at the bottom as well as change the label offset, now easily possible)
When not overriding these new methods it will do the same as before.

A corresponding pull-request for ios-charts has already been submitted by my colleague.
https://github.com/danielgindi/ios-charts/pull/432

Thank you very much!
Best Regards
Ben


PhilJay:
`LimitLine` now also extends `ComponentBase` and allows custom offsets.

forkch:
Thanks for implementing the label offset 

In my Pull-Request there was also a refactoring included which made it possible to change how a limit line is actually rendererd. By extracting the rendering of one limit line (in XAxisRenderer) into its own method one can then override this method to provide a custom rendering of the limit line.

PR #1072, opened 2015-10-05
Fix disappearing line when zooming into LineChart

cdeange:
If you have two values that are at separated by a large number of empty x-vals in between and you pinch to zoom inwards, eventually the line between them is not drawn. This is because the two `DataSet.getEntryForXIndex` calls may return the same `Entry`. The fix is to try and modify the min and max values to accommodate drawing a single line, rather than no line.


cdeange:
Thanks @PhilJay! Any idea when v2.1.5 will be prepared/pushed to maven?

PR #1086, opened 2015-10-08
Fixing overlapping label closes #1085

### LINKED ISSUE #1085: X-axis labels are overlapping
When x-axis labels are long, they get overlapped and it's hard to read which one belongs to which bar.

sales-i:
Fixing bug that caused x-axis labels to overlap by introducing initial label truncation and gradually revealing more text as the user zooms in. closes #1085


sales-i:
Any particular reason for rejection?

PR #1142, opened 2015-10-23
Fixed the avoidFirstLastClipping feature

danielgindi:


PhilJay:
Thx, finally a real fix :-)

danielgindi:
Yeah :-)
One day we may make all labels act like that, instead of being abruptly clipped...

PR #1143, opened 2015-10-23
Moved highlightTouch logic to highlightValue & deprecated highlightTouch

danielgindi:
It's not necessarily highlighting a "touch", it receives a highlight object. Times have changed :-)

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/Chart.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/Chart.java
@@ -572,14 +572,12 @@ public abstract class Chart<T extends ChartData<? extends DataSet<? extends Entr
 
     /**
      * Highlights the values represented by the provided Highlight object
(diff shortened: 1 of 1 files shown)


MPChartLib/src/com/github/mikephil/charting/charts/Chart.java
```diff
@@ -572,14 +572,12 @@ public void highlightValue(int xIndex, int dataSetIndex) {
...
+     * This DOES NOT generate a callback to the OnChartValueSelectedListener.
      *
      * @param highlight contains information about which entry should be highlighted
      */
     public void highlightValue(Highlight highlight) {
-        if (highlight == null)
-            highlightValues(null);
-        else
-            highlightValues(new Highlight[]{highlight});
+        highlightValue(highlight);
```
danielgindi: @PhilJay  Oh man I forgot to add a `, false` in there! Could you do a quick fix?

PR #1179, opened 2015-11-04
Set image format and subdirectory when saving to gallery

Adric:


PhilJay:
Did you test if your code is working?

Adric:
I did, it works with all the code samples plus my own. Actually also fixes
a bug: save to gallery wasn't appending a file extension in some samples.

-- Adric

PR #1200, opened 2015-11-14
improve option setDrawValueAboveBar(false) for HorizontalBarChart.

Sash0k:
Draw labels from left side. See related screenshots.
before
after


Sash0k:
want to send 9-patch as second pull request, but add to it. Need to fix

PR #1201, opened 2015-11-14
improve option setDrawValueAboveBar(false) for HorizontalBarChart.

Sash0k:
Draw labels from left side.
See related screenshots.
before
after


AngelOfMors:
UPD.
In HorizontalBarChartRenderer need change:
isInBoundsX -> isInBoundsLeft.
When the width of the full Bar - not visibled value

AngelOfMors:
Don't merge?

PR #1235, opened 2015-11-27
Add bitmap in  data of an Entry to show Marker in BarChart and LineChart

Joxad:
Hi,

For my project, I needed the possibility to add multiple Marker on different charts.
I began with MarkerViews/HighLights, but I found it easier to add it in the Data of the Entry an object of type Bitmap.
I have not been able to use MarkerView, (with its method draw), so i cheated using Bitmap. 
In this case, there is a test in the renderer (bar char/line) to draw the bitmap.

For example  : 

    Intensity intensity = intensities.get(i);
                if (intensity.showAlert()) {
                    entries.add(new Entry(intensity.getValue(), i, BitmapFactory.decodeResource(getResources(), R.drawable.star)));
                } else {
                    entries.add(new Entry(intensity.getValue(), i));
                }

If you want a specific view, I create a CustomView then I transform it to a bitmap
 

    if (i % 3 == 0) {
                    CustomValueView customValueView = CustomValueView_.build(getContext());
                    customValueView.bind(value);
                    entry = new BarEntry(value, i, customValueView.toBitmap());
                } else {
                    entry = new BarEntry(value, i);
      }

The toBitmap() method :

        
    public static Bitmap getBitmapFromView(View v) {
            v.setDrawingCacheEnabled(true);
    
    // this is the important code :)
    // Without it the view will have a dimension of 0,0 and the bitmap will be null
            v.measure(View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED),
                    View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED));
            v.layout(0, 0, v.getMeasuredWidth(), v.getMeasuredHeight());
    
            v.buildDrawingCache(true);
            Bitmap b = Bitmap.createBitmap(v.getDrawingCache());
            v.setDrawingCacheEnabled(false); // clear drawing cache
            return b;
        }

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/data/Entry.java
+++ b/MPChartLib/src/com/github/mikephil/charting/data/Entry.java
@@ -25,9 +25,6 @@ public class Entry implements Parcelable {
     private Object mData = null;
 
 
-    /** the marker view associated to the highlight */
-    private MarkerView markerView;
-
     /**
      * A Entry represents one single entry in the chart.
      * 
@@ -41,19 +38,7 @@ public class Entry implements Parcelable {
         mXIndex = xIndex;
     }
 
-    /**
-     * A Entry represents one single entry in the chart.
-     *
-     * @param val the y value (the actual value of the entry)
-     * @param xIndex the corresponding index in the x value array (index on the
-     *            x-axis of the chart, must NOT be higher than the length of the
-     *            x-values String array)
-     */
-    public Entry(float val, int xIndex, MarkerView markerView) {
-        mVal = val;
-        mXIndex = xIndex;
-        this.markerView = markerView;
-    }
+
     /**
      * A Entry represents one single entry in the chart.
      * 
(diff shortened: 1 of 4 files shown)


PhilJay:
Hello and thank you for your Pull Request.

However, for the follwing reason, I will not merge it:

 - When you propose a PR to a public library project, always bear in mind that the library might be used by a large variety of different people, with different use cases
 - You added a feature to directly draw entry data when it is a `Bitmap`. There are (or there might be) people who use this data object for simply storing a bitmap (and retrieval at later point), but do not want to draw it. If I merge your pull request, a lot of people might get unexpected behaviour because suddenly the `Bitmap` they added is drawn onto their chart.
 - You provided no methods for configuration. It is not possible to disable the feature you added.
 - Your PR is incomplete (and inconsistent), because all entry objects support storing "data" in them, and all chart types allow drawing the `MarkerView`, however, you only implemented the drawing feature for line and barchart.

PR #1257, opened 2015-12-08
Gradle project structure

JohnFrontzos:
Configure project directories to use default Android Gradle structure. Delete unnecessary v4 support lib, clean up Android Manifest and move launcher icons to mipmap folders.


jaredsburrows:
:+1:

PR #1290, opened 2015-12-16
Removed ambiguity of mDataNotSet

danielgindi:
We can just check for mData being null,
knowing that this is the situation defined as "no data"


danielgindi:
https://github.com/danielgindi/ios-charts/issues/585

PR #1292, opened 2015-12-17
calcYValueSum should not be ABS

danielgindi:


PhilJay:
Btw, we removed this from the realm-branch already.
This code should only be in the `PieData`.
Because the other data objects actually dont need it.

danielgindi:
Yeah I know :)

PR #1322, opened 2015-12-30
Pie animation should be applied after rotationAngle

mikemonteith:
I think that the pie chart's rotationAngle should not be multiplied by the mAnimation.getPhaseY(). This means that animated segments grow outwards from the rotationAngle rather than always from 0 degrees.
Do you agree?


PhilJay:
I'll test it, lemme check

danielgindi:
@PhilJay did you test it thoroughly? It seems weird. We do have different meanings for X and Y phases for Pie animation...

Also there's this `0` angle now dangling in the code.

danielgindi:
And this code still applies phaseY to the sweep angle, not the slice size, so it's pretty inconsistent.

danielgindi:
Ok now I get what happened there. Never mind :-)
It's much better than before. I think people will appreciate it!

PR #1326, opened 2016-01-03
Text lineheight fix

danielgindi:
Fixes for offset in line rendered through the rotation helper


PhilJay:
The x-axis labels and the legend (if on bottom) seem to be pretty close together after this. Or was this already and issue before? Have you noticed?

danielgindi:
It looked the same before, only the labels of the x-axis were not vertically aligned correctly when *not* rotated...

People can play with the offsets in any case..

PR #1350, opened 2016-01-08
Prevent the matrix from reaching invalid values

danielgindi:
https://github.com/danielgindi/ios-charts/commit/9942aee9c424ba027e1e62079232d3dc5dcdf772


PhilJay:
Shouldn't you set scale to 1? Instead of 0?
I'm not sure :-)

danielgindi:
Well I'm not sure about `1` as that would mean "keep the matrix the same", while the data changed to something invalid or to "no data".
We do want, I think, to scale the axis and stuff to zero in "invalid" cases- and we have encountered several different cases where data could be invalid, possibly because of mistakes of users, mistakes in in input data etc.

PR #1371, opened 2016-01-14
fixes #1345: bug in drawCubic inside LineCharts

### LINKED ISSUE #1345: Bug in drawCubic?
Hey, first of all thanks for this cool library!

i think I found a bug inside the drawing mechanism of the lines, i suppose.

The first image shows a a line connecting the points in the linear style. There everything is fine.
screen shot 2016-01-07 at 15 15 58

The next image below shows the same graph, but with `drawCubic` set to `true`.
screen shot 2016-01-07 at 15 14 42

As you can see, there is a small line before the first point which is too long.
Maybe the starting point of the cubicLine inside the LineChartRenderer is wrong?

Thanks a lot!

maksim88:
see bug description for details and images


maksim88:
Just to be clear: this PR fixes a usecase where the staring point is moved and does not start at the start of y-axis.
The first picture is the current state in v 2.2.0 (your current)
mp-v2 2 0



The second picture shows the state after PR
fixed-pr

PR #1392, opened 2016-01-21
add line animation

zawn:


zawn:
Do you think that without the animation between points

PR #1454, opened 2016-02-09
Scatter - support for shape holes, new X shape

danielgindi:
This is for the next version, let it sink in for a few days :-)

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/ScatterChart.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/ScatterChart.java
@@ -18,12 +18,10 @@ import com.github.mikephil.charting.renderer.ScatterChartRenderer;
 public class ScatterChart extends BarLineChartBase<ScatterData> implements ScatterDataProvider {
 
     /**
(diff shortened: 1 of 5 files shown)


PhilJay:
Version 2.2.2? or 2.2.3?

danielgindi:
2.2.2 :-) We've ended up making the latest realm fixes in 2.2.2 and not needing 2.2.3, so it should be 2.2.2

danielgindi:
I think this can be merged now

PhilJay:
Quick question: What does a hole in a cross look like? :-)

danielgindi:
No idea. That's why crosses do not have holes ;-)
If you find out- let me know! lol

PhilJay:
Ok :-D

PhilJay:
Offtopic: Should I merge the branch where I removet the startatzero property?
Or do you think this is not a  good idea?

danielgindi:
If I read the code correctly - it is backwards compatible, adjusting the mixAxis accordingly right?
In this case I don't see any problem, except maybe that it is not that clear to people that they can use NaN values in floats... It would work better in Swift, where values can be nullable so its clear that the minimum is optional.

PR #1465, opened 2016-02-11
Candle sticks improvements and new features

danielgindi:
* Support for candles without the bars (just ticks)
* Support for a separate neutral color
* Fixed rendering so that shadows to not show through the bars
* Some performance enhancements


danielgindi:
Note that I removed the Buffer classes for candle drawing - as they it much harder maintaining the drawing code. Here I still use a preallocated buffer - but local, and with less processing (iterating once instead of twice over all data).

danielgindi:
I think we need to update the candlestick screenshot :-)

PR #1483, opened 2016-02-18
Improvements for drawing pie slices

danielgindi:
Xfer tricks not needed anymore - we only draw what we need!

That `PorterDuffXfer` thing just erases stuff. It does not really solve things.
Now we draw what we need so we don't need tricks.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/PieChart.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/PieChart.java
@@ -16,6 +16,7 @@ import com.github.mikephil.charting.data.PieData;
 import com.github.mikephil.charting.highlight.Highlight;
 import com.github.mikephil.charting.interfaces.datasets.IPieDataSet;
 import com.github.mikephil.charting.renderer.PieChartRenderer;
+import com.github.mikephil.charting.utils.ColorTemplate;
 import com.github.mikephil.charting.utils.Utils;
 
 import java.util.List;
@@ -52,6 +53,11 @@ public class PieChart extends PieRadarChartBase<PieData> {
      * if true, the white hole inside the chart will be drawn
      */
     private boolean mDrawHole = true;
+
+    /**
+     * if true, the hole will see-through to the inner ends of the slices
+     */
+    private boolean mDrawHoleTransparent = true;
 
     /**
      * if true, the values inside the piechart are drawn as percent values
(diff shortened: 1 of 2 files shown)


PhilJay:
Unfortunately this induces the same problem as before the fix. Also, the hole that is supposed to be transparent is actually WHITE and not transparent.

screenshot_2016-02-18-18-19-43

danielgindi:
Are you sure you are testing the right branch?
Because on my device it works great :)
The hole allows solid colors and transparent colors, as well as the "transparentCircle"...

I'll test also to see that the branch was pushed correctly

PhilJay:
Okay, I`ll test again too :-)

PhilJay:
Just tested again. On my OnePlus One, the issue is still present

danielgindi:
Yes it does not work when there's no inner-radius. I need to push the inner corners and keep them sharp... I'll figure out the relation between the number of slices and the distance to push, and update the Pr :-)

danielgindi:
@PhilJay It seems like the stroke method has another problem!
When the slice is small and the stroke is wide - the stroke overflows to the other side...

screen shot 2016-02-19 at 1 45 19 pm

I've asked a question about this in math-exchange...
http://math.stackexchange.com/questions/1662812/calculating-distance-from-pie-center-for-correct-slice-spacing

PhilJay:
Okay, thanks for the effort!
I hope we can make this work, I would also perefer your solution.

danielgindi:
Done!
I ended up tearing the problem down like in geometry lessons in school, breaking into smaller shaped and smaller formulas until I figured out how to define the problem.

PhilJay:
Awesome, I just got home, will merge it tomorrow at the latest!

PhilJay:
Merged it. The issue related to "un-equal" slice space is gone!

However if I set the hole to be transparent `chart.setHoleColorTransparent(true)` it's still drawn in white. Any idea?

danielgindi:
Well first there's a hole color that you need to set, maybe to something with transparency. Like 0x50ffffff, is a semi-transparent white.
Also setting `chart.setHoleColorTransparent(false)` will tell it that we do not expect the color to be transparent, which means that we will not try to expand the center point into inner arcs as "we don't care what's behind the hole color as it's not transparent". So this will end up with full pie-slices, and the transparent hole color over it!

I've implemented it this way in iOS to begin with, to keep compatibility with the effect that is produces on Android with most combination of flags and colors...

PhilJay:
I see. So if I set the hole color to transparent it is actually transparent (I just tested that). But the transparent circle which slightly covers the pie-slices is also not drawn in that case.

So there is actually no bug in the current implementation, however I would change a few things:

 - Remove the `setHoleColorTransparent(...)` method. If we have to set the hole color transparent to "activate" this then why is it needed?
 - The transparent circle that (partly) covers the pie slices should be independent of the hole color. Meaning that it should also be drawn if the hole color is set to transparent, since there is a different color setter method for the transparent circle.

What do you think?

danielgindi:
I think you are right in both cases. Now that we have a unified implementation - where we draw what we want and do not rely on hacks - then we can actually put order to things.

But we do need to allow the same effect of having full pie slices - with the hole (and the "transparent" circle) over them. That's just about all that `setHoleColorTransparent` is doing, so we need a new name for it, maybe flip its value, and keep the original name marked as `@Deprecated`, calling the new one.

We should also mention in comments in what version a deprecated method is expected to be completely removed, and this way we can get rid of old stuff without breaking things without warning :-)

PhilJay:
Okay, have you already started working on this?
If not, I'll do it, just let me know.

danielgindi:
Nope, I'm on other stuff, you can go ahead :-)
Basically work to be done is:

1. "transparent circle" - always draw if properties are specified
2. "setHoleColorTransparent" - change to something like "setHoleCutsSlices" or "setHoleHidesSlices" or something like that.
3. Maybe the transparent circle should get a new name also?
4. Backwards compatibility :-)

michel-t-86:
`setHoleColorTransparent(...)` has not been set to `@Deprecated` but taken out all together.

I thought the idea would be to have it deprecated and internally call `setDrawSlicesUnderHole(...)`

michel-t-86:
Like was done for `setCircleSize(...)` for example.

If you guys agree I'll implement that and do a pull request?

danielgindi:
Go ahead :)

Issue #1507, opened 2016-02-24
LimitLine Labels can't render multi lines using backslash n ("\n")


silva96:
When you do `limitline.setLabel("FOO\nBAR")` it gets rendered as a single line as the picture shows (red one RIGHT_BOTTOM of first red limitline)

![screenshot_20160224-135259](https://cloud.githubusercontent.com/assets/4019924/13293509/b6134684-dafe-11e5-8ca1-1c2a12d66710.png)

PhilJay:
Yes, this is related to the fact that the canvas drawText(...) method ignores linebreaks. I might consider adding such a feature in the future.

PR #1508, opened 2016-02-24
Multiline labels

silva96:
To solve issue #1507 added multi line support via \n in the label string, also added LineSpacing integer as member of LimitLine in order to separate each line of the label as desired. Also cleaned up the code (cmd + shift + L ) thats why there are so many small changes)

screenshot_20160224-135019
screenshot_20160224-135602

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/components/LimitLine.java
+++ b/MPChartLib/src/com/github/mikephil/charting/components/LimitLine.java
@@ -44,6 +44,26 @@ public class LimitLine extends ComponentBase {
     }
 
     /**
+     * Returns the Line spacing (number of pixels that separates each line of the label)
+     * @return
+     */
+    public int getLabelLineSpacing() {
+        return mLabelLineSpacing;
+    }
+    /**
+     * Sets the Line spacing (number of pixels that separates each line of the label)
+     * for this LimitLine's Label.
+     *
+     * @param spacing - number of pixels that separates each line of the label
+     */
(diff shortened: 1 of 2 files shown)


danielgindi:
There are better, built-in methods in Android to draw multiline text. 
There's actually a drawing function prepared for that in the Utils of MPAndroidCharts.

silva96:
@danielgindi I want to collaborate, can you please tell me where do I find this drawing function? which built-in methods do you have in mind?

Best.

rstuven:
@danielgindi @PhilJay Functions in Utils are more for line breaking given a fixed width (constrainedToSize.width). Perhaps this pull request's name should be more specific ("Forced multiline labels using \n"), but still is a needed feature when the target width is unknown.

danielgindi:
Those functions support newlines as well, and use the OS's layout functions
to lay them out, no synthetic line heights etc.
Even when word wrapping is not desired- those functions should be used,
just without a width constraint.
בתאריך יום ו׳, 11 במרץ 2016 ב-16:43 מאת Ricardo Stuven <
notifications@github.com>:

c-castillo:
+1 to this feature

nkittsteiner:
+1. I need this feature.

danielgindi:
Take a look at `drawMultilineText` in Utils.

silva96:
I know, we proposed a new pull request using it. it's #1570 , take a look @danielgindi

PR #1565, opened 2016-03-10
Keep position on rotation

leoMehlig:
I’ve added a flag (keepPositionOnRotation) to BarLineChartBase
indicating whether the chart
should stay at its current position or not.

This is the same feature as described in the PR on [ios-charts](https://github.com/danielgindi/ios-charts/pull/824)

In the examples this feature doesn't work properly, because the Activities get reloaded on every rotation. This can be turned of be adding `    <activity android:name="BarChartActivity"
            android:configChanges="keyboardHidden|orientation|screenSize"></activity>` in the AppManifest.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -107,6 +107,11 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
     protected float mMinOffset = 15.f;
 
     /**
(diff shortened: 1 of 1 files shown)


silva96:
+1 to this, does it work for linechart too?

leoMehlig:
@silva96 Works for all Charts that inherit from BarLineChartBase. Basically all charts that move.

danielgindi:
Done beautifully on the first take, conforming to the style of the framework on both Android and iOS. Thank you!

PhilJay:
Currently only on my mobile, I will review it later.
But if Daniel says it's beautiful, I guess it really is :-)

PR #1570, opened 2016-03-11
Adding multiline support via backslash n for limitline labels 

silva96:
using Utils.drawMultilineText

This is an amend for pull request #1508


silva96:
Why closes it?

PR #1610, opened 2016-03-23
Fix crash with null Range when MarkerView is set and BarEntry has only one value

kumien:
Hi.
I have faced a problem when BarEntry has was created with float array but with only one element. When I added MarkerView I received NullPointerException from library.

> E/AndroidRuntime: FATAL EXCEPTION: main Process: com.zoetis.uk.zplay.vet, PID: 25639 java.lang.NullPointerException: Attempt to read from field 'float com.github.mikephil.charting.highlight.Range.to' on a null object reference
at com.github.mikephil.charting.charts.BarLineChartBase.getMarkerPosition(BarLineChartBase.java:582)
at com.github.mikephil.charting.charts.Chart.drawMarkers(Chart.java:681)
at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(BarLineChartBase.java:282)
at android.view.View.draw(View.java:16206)

I used BarChart with stacked values but sometimes I am receiving only one series and I faced with this crash. Thanks to this fix BarChart won't recognize BarEntry with one element array as stacked type.

**Second improvement.**
I am using LineBarChart to present weight and I want to show correct time on X axis. That's why I added x indexes without value. Everything worked correctly until I had to highlight values. LineBarChart uses all x indexes and if there wasn't values MarkerView didn't show up. Showing MarkerView was very hard because user has to click exactly on x index with value. I did improvement in this case and developer can enable searching x indexes with values for highlighting. In comment I also added information that it it is searching at runtime and we can observe problems with performance.

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -80,6 +80,12 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
     protected boolean mHighlightPerDragEnabled = true;
 
     /**
+     * Flag that indicates if highlighting will be enabled for x index with values.
+     * INFORMATION: If enabled then chart will do some calculations at runtime. Performance issues can occur.
+     */
+    protected boolean mHighlightXIndexWithValues = false;
+
+    /**
      * if true, dragging is enabled for the chart
      */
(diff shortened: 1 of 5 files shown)


johngray1965:
I'm seeing this crash too.  My case is a BarChart with stacked values.  Works if I pass in a single float (not as an array, but the api that takes a single float, ie not stacked).  But it crashes when I pass an array of two floats (stacked), and show a MarkerView.

PR #1670, opened 2016-04-10
Bar border

danielgindi:


StephenVinouze:
Hi @danielgindi,

I am extremely pleased to see you recently implemented this feature recently! I have just this requirement and planned to implement it by myself so you can guess my relief :)

Just a quick question though: is it by design that the border width is to be set in pixel and not in dp as it is mostly the case in the whole library (which will be properly converted in the library using the `Utils.convertDpToPixel(width)` method) ?

Thanks again for making our life easier!
Stephen

danielgindi:
That was just an oversight :-) Fixed now!

StephenVinouze:
Excellent, thanks for the quick reply!

danielgindi:
You're welcome :-)

PR #1682, opened 2016-04-13
potential solution to rare out of memory problems

brcinho:
The test were performed by manually opening and closing line chart (or pie chart) as fast as possible in a loop

Results before changes:

before_changes

Results after changes:

after_changes

I don't know which code change contributed more, setting canvas to null or calling gc

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/renderer/PieChartRenderer.java
+++ b/MPChartLib/src/com/github/mikephil/charting/renderer/PieChartRenderer.java
@@ -919,10 +919,15 @@ public class PieChartRenderer extends DataRenderer {
      * Releases the drawing bitmap. This should be called when {@link LineChart#onDetachedFromWindow()}.
      */
     public void releaseBitmap() {
+        if (mBitmapCanvas != null) {
+            mBitmapCanvas.setBitmap(null);
+            mBitmapCanvas = null;
+        }
         if (mDrawBitmap != null) {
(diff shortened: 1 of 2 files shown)


PhilJay:
Thanks, I'll test it tomorrow!
Have you observed any differences when calling gc() and not calling it?

brcinho:
NP. I'll repeat the tests tomorrow, do some more comparison and let you
know.

MPChartLib/src/com/github/mikephil/charting/renderer/PieChartRenderer.java
```diff
@@ -919,10 +919,15 @@ protected void drawRoundedSlices(Canvas c) {
...
+        if (mBitmapCanvas != null) {
+            mBitmapCanvas.setBitmap(null);
+            mBitmapCanvas = null;
+        }
         if (mDrawBitmap != null) {
             mDrawBitmap.get().recycle();
             mDrawBitmap.clear();
             mDrawBitmap = null;
         }
+        System.gc();
```
jaredsburrows: This does not cause performance issues? http://stackoverflow.com/questions/2414105/why-is-it-bad-practice-to-call-system-gc

jaredsburrows:
:+1:

brcinho:
Seems that System.gc() call was not necessary. Setting canvas to null and
settings its reference to bitmap to null solved the problem I think.

PR #1744, opened 2016-05-05
LargeValueFormatter

AngelOfMors:
The other language need more characters for suffix


AngelOfMors:
The other language need more characters for suffix (two and more)

PhilJay:
Will be fixed in v3.0.0

PR #1775, opened 2016-05-13
Can associate form to label for each data set

StephenVinouze:
Hi @PhilJay ,

I added the ability to draw form next to each x-value label that can be easily customized with size, radius (0 by default which draws a square but can be used to have a rounded square or a circle depending of the radius value). You can also set a margin between the form and the label with an `offset` parameter.

Both form and label combined are vertically and horizontally aligned depending of their size, offset, etc... and can be done in each axis position. Only the horizontal chart was not implemented (I believe this is not a wanted feature).

Of course, associated forms with label is optional. You can even decide to draw a form for only a specific set of entries.

Setting forms is done the same way as x-values but through a setter in the `ChartData` class (no constructor created with this) and its array size must match the x-values array size (I extended the behavior of the `checkLegal()` method). If you need to draw forms for only some entries you just need to add a `ColorTemplate.COLOR_SKIP` to the array and it will be ignored. The customization of the form is done through properties in the `XAxis` class.

I am aware this is not a "must-have" feature but I came across this requirement twice in a row to highlight some specific entries in the X axis labels so I figured it would be a "could-have" feature for this library ;)

Let me know how you feel about it. As always you will find a screenshot attached below to illustrate this feature ;)

Best regards,
Stephen

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -15,8 +15,6 @@ import android.util.AttributeSet;
 import android.util.Log;
 import android.view.MotionEvent;
 
-import com.github.mikephil.charting.components.Legend;
-import com.github.mikephil.charting.components.Legend.LegendPosition;
 import com.github.mikephil.charting.components.XAxis.XAxisPosition;
 import com.github.mikephil.charting.components.YAxis;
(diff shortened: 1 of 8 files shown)


StephenVinouze:
axis_form

almic:
There have been some changes to the repo since this pull request, I encourage you to check if the current version does this already or not, and submit a new pull request if you want. Otherwise, feel free to search open issues and if this isn't mentioned, then please open an issue describing what problem this fixes.

Thank you!

PR #1875, opened 2016-06-15
Custom shape renderer - easy to implement - Optemised

wajdichamakhi:
Changing ScatterShape from ENUM to HashMap of String .
Creating a provider which provides ShapeRenderer.
Creating CustomShapeRenderers imlementing ShapeRender with different renderShape() method
Replacing the Switch Statement by ShapeRenderer.renderShape();

=> Better OOP implementation. Scalable easily.
To add new Shape you only need to create a class implementing ShapeRenderer and override renderShape method and then add the class in the provider ScatterChart.registerShapeRenderer(String scatterShapeName, ShapeRenderer shapeRenderer).

### DIFF
--- a/MPChartExample/build.gradle
+++ b/MPChartExample/build.gradle
@@ -37,7 +37,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.0'
+        classpath 'com.android.tools.build:gradle:2.1.2'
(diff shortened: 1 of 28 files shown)


PhilJay:
Thanks! Could your remove the changes you made regarding the "bold" text? Other than that, this looks good.

wajdichamakhi:
I created a new pull request for custom Shapes with no bold control

Issue #1897, opened 2016-06-22
Reduce slow draw calls in Realtime graph by caching drawn circles as bitmaps, then drawing the bitmaps directly to canvas.


tonypatino-monoclesociety:
Measuring the performance of the application using gfxinfo, I note the following performance on a Nexus 6.  Upon investigating which methods were consuming the most time, I found a culprit in lineChartRenderer.drawCircles.  The repeated calls to canvas.drawPath and canvas.drawCircle result in many slow draw commands as noted below, and lead to a worst case rendering time of 73ms.

Total frames rendered: 701
Janky frames: 480 (68.47%)
90th percentile: 42ms
95th percentile: 53ms
99th percentile: 73ms
Number Missed Vsync: 5
Number High input latency: 0
Number Slow UI thread: 278
Number Slow bitmap uploads: 401
Number Slow issue draw commands: 193

By caching the circles as bitmaps, it's possible to eliminate almost all of the slow draw calls, at the cost of spending slightly more time on the UI thread, thereby cutting the rendering time of all three measured bottom percentile frames by 25% - 40%.

Total frames rendered: 743
Janky frames: 499 (67.16%)
90th percentile: 32ms
95th percentile: 34ms
99th percentile: 44ms
Number Missed Vsync: 3
Number High input latency: 0
Number Slow UI thread: 445
Number Slow bitmap uploads: 490
Number Slow issue draw commands: 9

The amount of time spent on the UI thread is still troublesome, at an average of 9ms per frame according to the chart when mChart.setLogEnabled(true), compared to the master branch's 8ms per frame.  Perhaps some of this work can be pre-calculated ahead of time, or pushed to a worker thread. in order to smooth out and reduce the framerate further.

tonypatino-monoclesociety:
Pull request #1908
Attached to issue #1892

Addresses this issue as well.

PR #1899, opened 2016-06-22
Cache circles in LineChartRenderer instead of recreating them (#1897)

tonypatino-monoclesociety:
IDataSet now has a UserData object which can store caches of data that may be expensive to calculate on-the-fly.  In this case, LineChartRenderer defines a UserData object to pre-draw and cache the circles that would be drawn inside of drawCircles().  Since canvas.drawBitmap() offers significant savings over the previous canvas.drawCircle()/canvas.drawPath(), we reduce the worst case rendering times by up to 40%.  This savings is noticeable in the RealtimeLineChartActivity.  Said activity is updated to contain many different colors, to ensure that the bitmap caching properly and correctly paints the circles as expected.

See ticket 1897 : https://github.com/PhilJay/MPAndroidChart/issues/1897

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/interfaces/datasets/IDataSet.java
+++ b/MPChartLib/src/com/github/mikephil/charting/interfaces/datasets/IDataSet.java
@@ -393,15 +393,4 @@ public interface IDataSet<T extends Entry> {
      */
     boolean isVisible();
 
-    /**
-     * Returns miscellaneous data stored by this IDataSet instance.
-     *
-     * @return
-     */
-    Object getUserData();
-
-    /**
-     * Set miscellaneous data in this IDataSet instance.
-     */
-    void setUserData(Object value);
 }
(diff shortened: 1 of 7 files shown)


MPChartLib/src/com/github/mikephil/charting/interfaces/datasets/IDataSet.java
```diff
@@ -392,4 +392,16 @@
      * @return
      */
     boolean isVisible();
+
```
tonypatino-monoclesociety: As I think about it, perhaps it would be better to call this object ImageCache rather than UserData, since that would prevent confusion regarding when or where the object is intended to be used.

tonypatino-monoclesociety:
_Pull request #1908 would make this redundant, as all changes in this pull request are present there as well._

tonypatino-monoclesociety:
Pull request #1908 would make this redundant, as all changes in this pull request are present there as well.

Issue #1892, opened 2016-06-20
Eliminate GC-induced stuttering during animations by eliminating wasteful allocations


tonypatino-monoclesociety:
I noticed that the Realtime chart demo can stutter due to many repeated and wasteful allocations in sections of code that are called in each update of the data, and it would be great to smooth out the animations and remove stuttering.  I've so far identified, verified, and tested ways to reduce the number of allocations by approximately 70% and memory consumed by about 30%, with an attendant improvement in animation quality through various means, such as:

1. Eliminating the use of convenience iterations in favor of old fashioned iteration.
for(;;){}  instead of
for(Object foo : inSet){} 

2. Caching the results of String Formatting and using those cached results when the value in the dataset has not changed

3. Using field level references to store short-lived instances within methods and re-setting them instead of instantiating a new one
  eg 1: instead of instantiating a float[] temp = new float[2]; inside a method
declare the float[] as a private field and initialize its two locations as needed.
  eg 2: instead of instantiating a new Matrix inside of a function, create a Matrix field and call matrix.reset(); where previously a new Matrix would be instantiated.

4. Pre-instantiating runnables outside of a local Thread object's run method when each Runnable instance performs the same action on the same object instance

Other means that I have not yet explored but deserve some research include using object pooling for frequently instantiated objects such as FSize, PointF, and the like.  

Attached are the results from the memory monitor in Android Studio

[allocationsComparison.zip](https://github.com/PhilJay/MPAndroidChart/files/324054/allocationsComparison.zip)

PhilJay:
Thanks for the input, I know there is definitely some space upwards regarding performance.

I am currently crazy busy with other optimizations of the library regarding v3.0 release, so if you could point me to the exact areas in code where the most speed is "left behind" that would be great and it could save me some time when I work on the changes. Thanks!

tonypatino-monoclesociety:
No worries about the schedule, Phil.  I've been considering various ways to up the performance in the library, and they aren't all fully tested yet, but I'd be happy to create a pull request with my findings so far.    The largest gain I found was related to #1897 which there's a pull request for.  

The biggest hurdle at this moment is the amount of work performed on the UI thread.  Since much of the calculation is done inside of onDraw instead of pre-calculated in a worker thread, things can look a little stuttery from time to time.  

[...]

PR #1908, opened 2016-06-24
Eliminate allocs, cache images/text (#1892)

tonypatino-monoclesociety:
As detailed in ticket 1892, this change set implements many changes to reduce allocations during Realtime chart rendering.  Among the changes:

1. ValueFormatters getFormattedValue property now accepts an int position property to facilitate caching of formatted strings.

2. Many methods that instantiated temp variables now rely on instance-level fields instead, and only instantiate once.  Temp variables are reset instead of reinstantiated.

3. ObjectPool implemented for frequently instantiated value objects, such as PointD and FSize.
    NOTE: FSize is no longer truly immutable, but its fields are encapsulated with getters-only.  This is to facilitate pooling.  Field access would have faster performance, but I wanted to err on the side of keeping the class read-only as far as owners of an instance are concerned.

4. LineDataSet offers a getCircleColorCount() method, for use in caching of circle bitmaps.

5. Change convenience for:in arrays into for(;;) arrays.  Convenience array iteration has a memory cost, and allocates an object for each item in the List.

6. Caching of bitmap circles in the line chart renderer.  This makes ticket #1897 redundant (as well as its pull request).

### DIFF
--- a/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -527,6 +527,7 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
     /**
      * calculates the modulus for x-labels and grid
      */
+    float[] valuesForCalcModulus = new float[9];
     protected void calcModulus() {
 
         if (mXAxis == null || !mXAxis.isEnabled())
@@ -534,7 +535,10 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
 
         if (!mXAxis.isAxisModulusCustom()) {
 
-            float[] values = new float[9];
+            float[] values = valuesForCalcModulus;
+            for(int i = 0 ; i < values.length ; i++){
+                values[i] = 0;
+            }
             mViewPortHandler.getMatrixTouch().getValues(values);
 
             mXAxis.mAxisLabelModulus = (int) Math
@@ -1284,10 +1288,11 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
      * @param y
      * @return
      */
+    private float[] ptsForGetValuesByTouchPoint = new float[2];
     public PointD getValuesByTouchPoint(float x, float y, AxisDependency axis) {
 
         // create an array of the touch-point
-        float[] pts = new float[2];
+        float[] pts = ptsForGetValuesByTouchPoint;
         pts[0] = x;
         pts[1] = y;
 
@@ -1296,7 +1301,7 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
         double xTouchVal = pts[0];
         double yTouchVal = pts[1];
 
-        return new PointD(xTouchVal, yTouchVal);
+        return PointD.getInstance(xTouchVal, yTouchVal);
     }
 
     /**
@@ -1307,15 +1312,16 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
      * @param y
      * @return
      */
+    private float[] ptsForGetPixelsForValues = new float[2];
     public PointD getPixelsForValues(float x, float y, AxisDependency axis) {
 
-        float[] pts = new float[]{
-                x, y
-        };
+        float[] pts = ptsForGetPixelsForValues;
+        pts[0] = x;
+        pts[1] = y;
 
         getTransformer(axis).pointValuesToPixel(pts);
 
-        return new PointD(pts[0], pts[1]);
+        return PointD.getInstance(pts[0], pts[1]);
     }
 
     /**
@@ -1366,13 +1372,13 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
      *
      * @return
      */
+    private float[] lowestVisibleXIndexPoints = new float[2];
     @Override
     public int getLowestVisibleXIndex() {
-        float[] pts = new float[]{
-                mViewPortHandler.contentLeft(), mViewPortHandler.contentBottom()
-        };
-        getTransformer(AxisDependency.LEFT).pixelsToValue(pts);
-        return (pts[0] <= 0) ? 0 : (int)Math.ceil(pts[0]);
+        lowestVisibleXIndexPoints[0] = mViewPortHandler.contentLeft();
+        lowestVisibleXIndexPoints[1] = mViewPortHandler.contentBottom();
+        getTransformer(AxisDependency.LEFT).pixelsToValue(lowestVisibleXIndexPoints);
+        return (lowestVisibleXIndexPoints[0] <= 0) ? 0 : (int)Math.ceil(lowestVisibleXIndexPoints[0]);
     }
 
     /**
@@ -1381,13 +1387,13 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
      *
      * @return
      */
+    private float[] highestVisibleXIndexPoints = new float[2];
     @Override
     public int getHighestVisibleXIndex() {
-        float[] pts = new float[]{
-                mViewPortHandler.contentRight(), mViewPortHandler.contentBottom()
-        };
-        getTransformer(AxisDependency.LEFT).pixelsToValue(pts);
-        return Math.min(mData.getXValCount() - 1, (int)Math.floor(pts[0]));
+        highestVisibleXIndexPoints[0] = mViewPortHandler.contentRight();
+        highestVisibleXIndexPoints[1] = mViewPortHandler.contentBottom();
+        getTransformer(AxisDependency.LEFT).pixelsToValue(highestVisibleXIndexPoints);
+        return Math.min(mData.getXValCount() - 1, (int)Math.floor(highestVisibleXIndexPoints[0]));
     }
 
     /**
--- a/MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
+++ b/MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
@@ -233,7 +233,8 @@ public abstract class ChartData<T extends IDataSet<? extends Entry>> {
                 mLeftAxisMax = firstLeft.getYMax();
                 mLeftAxisMin = firstLeft.getYMin();
 
-                for (IDataSet dataSet : mDataSets) {
+                for (int i = 0 ; i < mDataSets.size() ; i++){
+                    IDataSet dataSet = mDataSets.get(i);
                     if (dataSet.getAxisDependency() == AxisDependency.LEFT) {
                         if (dataSet.getYMin() < mLeftAxisMin)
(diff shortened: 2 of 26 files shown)


PhilJay:
Thanks man, looks great!
Sorry that I forgot to mention this, but could you by any chance create this PR for the "remove_xindices" branch? This is the new branch for v3.0 and it's ~150 commits ahead of master.

It's code basis is very different from what is currently available on master so it would be a huge help if you could create the performance improvements for the new version 3.0 of the library.

tonypatino-monoclesociety:
Sure thing, I'll pull that branch today.  If it's not done today I'll have to resume the merge on Monday next week.

tonypatino-monoclesociety:
@PhilJay Do you have strong feelings about PointF vs PointD?  PointF can't extend Poolable and would need a wrapper of some sort, which would lead right back to PointD.

FabianTerhorst:
Since art jit isn´t in use anymore and without jit the current foreach implementation is the fastest. And your new implementation would try to get the arraylenght once on every iteration.

tonypatino-monoclesociety:
@FabianTerhorst Thanks for the input.  Could you post some memory consumption and performance figures for the master branch vs the feature branch to verify whether the old implementation is faster or more memory efficient?  My measurements show a sharp drop in worst-case frame rendering of up to 40% now.  

It would be easy enough to extract the array length check in any case.

PhilJay:
I'm not sure how much boost traditional for loops bring vs. foreach. But reducing the allocations will definitely boost performance.

FabianTerhorst:
The first step would be to assign the Array length to a Variable.

tonypatino-monoclesociety:
@PhilJay See the new pull request : https://github.com/PhilJay/MPAndroidChart/pull/1920

The new axis renderer in remove_xindeces presents an issue that prevents caching the labels in the same manner we could in v2.  I think that change should probably be its own separate issue in the issue tracker rather than part of this one.

MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
```diff
@@ -527,14 +527,18 @@ public void calculateOffsets() {
...
     protected void calcModulus() {
 
         if (mXAxis == null || !mXAxis.isEnabled())
             return;
 
         if (!mXAxis.isAxisModulusCustom()) {
 
-            float[] values = new float[9];
+            float[] values = valuesForCalcModulus;
+            for(int i = 0 ; i < values.length ; i++){
```
FabianTerhorst: like discussed:
```
int length = pts.length;
for(int i = 0 ; i < length ; i++)
```

MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
```diff
@@ -810,7 +812,8 @@ public int getIndexOfDataSet(T dataSet) {
      * @return
      */
     public T getFirstLeft() {
-        for (T dataSet : mDataSets) {
+        for(int i = 0 ; i < mDataSets.size() ; i++){
```
FabianTerhorst: also

MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
```diff
@@ -825,7 +828,8 @@ public T getFirstLeft() {
      * @return
      */
     public T getFirstRight() {
-        for (T dataSet : mDataSets) {
+        for(int i = 0 ; i < mDataSets.size() ; i++){
```
FabianTerhorst: also

MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
```diff
@@ -859,7 +863,8 @@ public void setValueFormatter(ValueFormatter f) {
         if (f == null)
             return;
         else {
-            for (IDataSet set : mDataSets) {
+            for (int i = 0 ; i < mDataSets.size() ; i++){
```
FabianTerhorst: +

MPChartLib/src/com/github/mikephil/charting/data/ChartData.java
```diff
@@ -872,7 +877,8 @@ public void setValueFormatter(ValueFormatter f) {
      * @param color
      */
     public void setValueTextColor(int color) {
-        for (IDataSet set : mDataSets) {
+        for (int i = 0 ; i < mDataSets.size() ; i++){
```
FabianTerhorst: +

MPChartLib/src/com/github/mikephil/charting/charts/BarLineChartBase.java
tonypatino-monoclesociety: This and all other similar items are addressed in my local branch based on the remove_xindeces branch.  remove_xindeces will soon be merged into master, and it introduces many changes that will make this pull redundant (eg: changes to the Axis renderers, completely removing Realm from the repo, etc).

tonypatino-monoclesociety:
Since remove_xindeces will be merged into Master soon, this pull will be somewhat redundant (various classes no longer exist in remove_xindeces, for example).

PhilJay:
I did some performance testing regarding loop iterations. Below you can see my test class and the results. Please tell me if there is something wrong with the test, but I do not see any significant benefit of a traditional for loop vs. the "optimized" for loop. However, there seems to be a benefit when using traditional for loops over foreach loops.

What has to be mentioned here though is the fact that neither of the loops do extensive work inside the iteration, so loop optimization in a "real" use case will lead to much lower performance increase.

```java

public class LoopPerformanceTest {

	public static final int TEN_MILLION = 10000000;
	public static final int ONE_MILLION = 1000000;
	
	public static final int RUNS = 100;

	public static void main(String[] args) {
		
		List<Integer> list = new ArrayList<Integer>();
		
		for(int i = 0; i < TEN_MILLION; i++) {
			list.add(1);
		}
		
		long time1Sum = 0;
		long time2Sum = 0;
		long time3Sum = 0;

		for(int i = 0; i < RUNS; i++) {
			time1Sum += traditionalLoop(list);
			time2Sum += traditionalLoopOptimized(list);
			time3Sum += foreachLoop(list);
		}
		
		System.out.println("Traditional loop: " + time1Sum / ONE_MILLION);
		System.out.println("Traditional loop optimized: " + time2Sum / ONE_MILLION);
		System.out.println("Foreach loop: " + time3Sum / ONE_MILLION);
	}

	public static long traditionalLoop(List<Integer> list) {

		long start = System.nanoTime();
		long sum = 0;

		for (int i = 0; i < list.size(); i++) {
			sum += list.get(i);
		}
		
		return System.nanoTime() - start;
	}

	public static long traditionalLoopOptimized(List<Integer> list) {

		long start = System.nanoTime();
		long sum = 0;
		int size = list.size();

		for (int i = 0; i < size; i++) {
			sum += list.get(i);
		}

		return System.nanoTime() - start;
	}
	
	public static long foreachLoop(List<Integer> list) {

		long start = System.nanoTime();
		long sum = 0;

		for (int i : list) {
			sum += i;
		}

		return System.nanoTime() - start;
	}
}
```

Results:

```java
Traditional loop: 806
Traditional loop optimized: 803
Foreach loop: 1019
```

tonypatino-monoclesociety:
Agreed that extracting the array size offers only a small benefit.  I wouldn't go out of my way to do it in the future, at least on Android, but the results of traditional loop vs foreach are indeed what I expected and what I've measured.  

Foreach has the added problem of implicitly instantiating a new Object for each iteration of the loop, which runs up the amount of garbage quickly when animating the charts, or adding new values.  

Honestly, the _biggest_ issue left for me to tackle in the proposed v3.0.0 release is how to port the ValueFormatter from this pull request into the remove_xindeces branch.  In v2 it was much easier to cache the Axis labels, but the changes in v3 make that a much bigger problem.  Storing array length outside a loop declaration is small potatoes compared to being unable to cache the results of formatted Strings, when looking at the amount of garbage the String formatting generates.

tonypatino-monoclesociety:
@PhilJay Got the ValueFormatter caching done.  The allocations on the v3 remove_xindeces branch are now down _even lower_ than they were in this pull request for v2 master.  I'll keep an eye out for your updates and make a pull request based off my v3 fork.

**Screenshot of the allocations in my fork tony/remove_xindeces** 
**0.39MB for 1000 iterations, 0 GC runs, no danger of one in the near future**
screen shot 2016-06-29 at 6 35 06 pm


**vs the current main remove_xindeces**
**4.8MB for 1000 iterations, 1 GC run with another impending quickly**
screen shot 2016-06-29 at 6 41 30 pm

PhilJay:
Awesome, thx!
Did you test the caching thoroughly? v3.0.0 will probably be good to go next week!

tonypatino-monoclesociety:
I have a unit test for the FormattedStringCache classes, as well as the ObjectPool class.  All existing unit tests continue to pass.  Nothing in the rendered charts looks wrong so far.

This change set will require a pull request on the Realm repo as well, since I added a method int getCircleColorCount() to the ILineDataSet interface.  One of the Realm classes needs to implement the method as well.

tonypatino-monoclesociety:
After replacing a few other [] allocations with List<> to [] copying, and hunting down a few other pooling opportunities --

*0.15MB Allocated memory over 1000 iterations of the Realtime chart, with only 2200 total invocations*

screen shot 2016-06-30 at 4 30 44 pm

PhilJay:
Thanks, thats awesome! Did you any testing on how these changes effect performance?
Should give some boost there as well I would guess.

tonypatino-monoclesociety:
The gfxinfo shows that all total, it helps reduce the worst-case render times by up to 40%.  Also, these changes make GC runs extremely infrequent which prevents garbage-related hiccups.  

I think that the next step would be to move all the bitmap rendering off the UI thread.  Drawing the maps is still a fairly expensive operation, and if it can happen outside of onDraw this should theoretically help.  There could be a double-buffered bitmap drawing scheme, such that onDraw will render Bitmap A, while the worker thread draws into Bitmap B.  Only once the worker is done, does Bitmap B get copied to Bitmap A.  

Also, it could be good to offer users a desired framerate.  Like, if they only want 30fps then only try to invalidate the display list every 33ms.

YuriDenison:
@PhilJay Please consider using JMH framework for all your JAVA perfomance testing.
You can not really trust benchmark you've created.

Here are more accurate numbers from my Macbook Air and JDK 8:
```
Benchmark                           Mode  Cnt    Score    Error   Units
LoopTest.foreachLoop               thrpt   10  280.758 ± 11.071  ops/ns
LoopTest.traditionalLoop           thrpt   10  320.631 ± 14.032  ops/ns
LoopTest.traditionalLoopOptimized  thrpt   10  311.157 ± 15.439  ops/ns 
```
[code](https://gist.github.com/YuriDenison/bca68914e69887cb8874ae51f18168b0)
[jmh](http://openjdk.java.net/projects/code-tools/jmh/)

Traditional and optimized versions are the same and faster than foreach.
In this simple case the results you got shows the same result but in any more complicated benchmark only JMH will show you results you can trust.

PR #1920, opened 2016-06-27
Eliminating allocs, caching values #1892

tonypatino-monoclesociety:
This ports work from the pull request #1908 to the new v3 version of the Libraries.  There remains one outstanding issue though, related to caching formatted strings for the XAxis.  A TODO has been noted for this.


tonypatino-monoclesociety:
@PhilJay the new v3 presents an interesting problem when considering caching formatted String values along the XAxis. Previously we were able to cache these values by associating the string with the specific Entry index in the DataSet, and only re-create the formatted text if the Entry value changed.

Since the new library doesn't do this, and all it tracks is the relative position of the label on the axis, we're forced to constantly re-create format strings in DefaultAxisValueFormatter as the axis changes in the Realtime sample.

The library should find a way to uniquely identify which entry an axis value is associated with so that the strings can be cached regardless of the X/Y position of the label along the Axis at render time.

FabianTerhorst:
Looks good to me. :+1:

tonypatino-monoclesociety:
Updated the request.  It appears files were missing.

tonypatino-monoclesociety:
Closing due to conflicts.  Will resolve them.

Issue #1895, opened 2016-06-21
Canvas.clipPath(...) crash when too many filled Entries


chintandesai49:
In Line chart when i increase first seekBar and when it reaches to approx value 350 app restarts it self.
I have attached zip file containing video of problem.
I tested this thing on Nexus 6. But not able to reproduc on nexus 4.
[MPAndroidChart Problem.zip](https://github.com/PhilJay/MPAndroidChart/files/324989/MPAndroidChart.Problem.zip)

PhilJay:
Thank you for the report. What kind of device are you using?

Could you tell me if the crash still occurs if you disable "filled" (the red gradient) option in the example you have shown me.

chintandesai49:
No crash at all just simple restart of app as you can see in device. I'm using nexus 6 and Nexus 4. 
problem occur in only Nexus 6 but not reproducible on Nexus 4.

PhilJay:
Can you please test again and see if the issue also occurs if "filled" is disabled?
The restart of the app is an indicator for a crash.

chintandesai49:
No, when toggle filled marked disable app does not restart and worked as expected on Nexus 6 and Nexus 4.

PhilJay:
Alright, this looks very much like a memory issue, most likely related to `canvas.clipPath(...)` method which is used for filling. If you disable hardware acceleration, the crash will stop.

tonypatino-monoclesociety:
@PhilJay I have a proposed fix for this bug currently in my local remove_xindeces branch.  It involves generating smaller paths iteratively until all paths are drawn.  Would you like to incorporate this after v3.0.0 is out?

tonypatino-monoclesociety:
**Results - Stress testing Activity 1 with 1500 items, full color fill.**
![device-2016-06-30-174125](https://cloud.githubusercontent.com/assets/13628883/16508655/f009a960-3ee9-11e6-887e-2aa835893d92.png)



**Results - Stress testing Activity 2 with 1500 items and full color fill.**
![device-2016-06-30-174419](https://cloud.githubusercontent.com/assets/13628883/16508674/332bbfa8-3eea-11e6-8270-9ba52684e0b5.png)

PhilJay:
Yes, I would incorporate it!

tonypatino-monoclesociety:
Cool, I'll incorporate this into the performance-related pull request.

PR #1947, opened 2016-07-01
Performance + Crash fix (#1892) (#1895)

tonypatino-monoclesociety:
This pull request contains --
1) All the allocation optimizations to the library, including unit tests for new classes.  (#1892)
2) Bitmap buffering of drawn circles in LineChartRenderer (#1892)
3) Crash fix and stress test for (#1895)

Note-- This change set will require approving a pull request in the MPAndroidChart-Realm project, to account for a new abstract method to retrieve the number of circle colors.  https://github.com/PhilJay/MPAndroidChart-Realm/pull/1

### DIFF
--- a/MPChartExample/res/layout/activity_linechart.xml
+++ b/MPChartExample/res/layout/activity_linechart.xml
@@ -18,7 +18,7 @@
         android:layout_margin="8dp"
         android:layout_toLeftOf="@+id/tvYMax"
         android:layout_marginRight="5dp"
-        android:max="200"
+        android:max="150"
         android:paddingBottom="12dp" />
 
     <SeekBar
@@ -30,7 +30,7 @@
         android:layout_marginBottom="35dp"
         android:layout_toLeftOf="@+id/tvXMax"
         android:layout_marginRight="5dp"
-        android:max="500"
+        android:max="1500"
         android:paddingBottom="12dp" />
 
     <TextView
--- a/MPChartExample/src/com/xxmassdeveloper/mpchartexample/BarChartActivity.java
+++ b/MPChartExample/src/com/xxmassdeveloper/mpchartexample/BarChartActivity.java
@@ -3,6 +3,7 @@ package com.xxmassdeveloper.mpchartexample;
 
 import android.annotation.SuppressLint;
 import android.graphics.PointF;
+import android.graphics.Rect;
 import android.graphics.RectF;
 import android.os.Bundle;
 import android.util.Log;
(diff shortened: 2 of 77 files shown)


danielgindi:
Is this actually *tested* and proved to be worth it?
As it can easily get out of hand - if someone forgets to recycle a size or a point, or returns too early etc. Also the overhead of taking an FSize from the pool and returning it - might sometimes be more costly than just allocating - or preallocating a single instance private to the class.

danielgindi:
Also all of that overhead just for the sake of pooling long-running objects like a formatter in the example, or an ZoomJob - that's just nonsense.

tonypatino-monoclesociety:
[comparison.zip](https://github.com/PhilJay/MPAndroidChart/files/404338/comparison.zip)

The two allocation charts show the app running at its default 500 allocation example.  Format string caching is a significant part of the memory allocation difference in these readings.  

The master branch incurred one GC run before the end of the 500 entries, and was near to incurring a second.  At 5000 entries, master incurred garbage collection about 15 times.  

The merged feature branch allocates a fraction of the memory.  Even at 5000 entries, it was not remotely close to incurring garbage collection.

The results from gfx info indicate that after all the changes the worst case screen rendering times improve between 25%-40%.  The rendering time per frame is much more consistent with these updates.  GC hiccups account for some of this difference.  The rest is accounted for by caching the the drawn circles inside the LineChartRenderer, which ends up far less expensive computationally than re-drawing all the circles during each onDraw() call.


As it stands, the end user neglecting to recycle an FSize or MPPointF doesn't result in issues beyond the original issue of wasted memory.

danielgindi:
How is caching the string formatter giving you any benefit? It does not make any sense. These are set once - and reused as a variable. There's no reason to cache them.

Well if the user neglecting to recycle an FSize or a point is not an issue - then it's logical to consider it not being an issue if not taken from the pool at all.

It seems like there's some good work here on the object pools, but in some cases it's just an overkill...

tonypatino-monoclesociety:
Caching the formatted string results prevents the application from calling
the mFormat.format() method repeatedly, to re-create formatted strings that
it has done the work to format once already.  The original code even
specifically calls DefaultAxisValueFormatter.getFormattedValue out as a
hotspot in the comments, indicating that memory allocation related to
.format() in the java codebase was a known issue

// avoid memory allocations here (for performance)

Unfortunately mFormat.format() is very allocation-hungry all on its own,
and the previous implementation would re-create format strings for values
that it had already once created.  The caching eliminates all excessive
allocations for strings that have already been generated, reducing garbage
substantially.

As for failing to recycle the points not being an issue in terms of correct
functioning of the application, the recycling exists as a way to improve
performance where it's needed, such as in animated graphs, or graphs that
update values over time.  It's there to allow users the option to be
economical.  The comparisons in the zip file are striking, and the option
is a good one to have.  The FSize and PointF savings in memory allocation
is larger still than the large savings from caching the strings, given the
thousands of instantiations that occur even with a small dataset.



*Tony Patino *| Development Lead
tonypatino@monoclesociety.com
206.718.8363 <+206-718-8363> (mobile)

PR #2079, opened 2016-07-29
Develop

AlainMuller:
Added radius attribute for barchart to display rounded bar charts (with rounded borders too)


montecillodavidacn:
Hi when will this be available in the release?

titiarimba:
Hi. is it available on v2.2.4 release?

Issue #2083, opened 2016-08-01
an IndexOutOfBoundsException occasionally occur  in my app linechart


Sevenupboy:
when i finish coding of my app ,    an IndexOutOfBoundsException occasionally  occur  as below ( open 20 times & occur once ) ,and when i make yAxis enable(false) the problem never occur....

> 08-01 16:08:53.360 20364-21189/com.qixi.m303chartdemo D/MainActivity: hum arrayHum.size():64
08-01 16:08:53.360 20364-20364/com.qixi.m303chartdemo D/AndroidRuntime: Shutting down VM
08-01 16:08:53.360 20364-21189/com.qixi.m303chartdemo D/MainActivity: init co2
08-01 16:08:53.360 20364-21189/com.qixi.m303chartdemo D/MainActivity: co2 setting
08-01 16:08:53.360 20364-21189/com.qixi.m303chartdemo D/MainActivity: co2 data3
08-01 16:08:53.360 20364-21189/com.qixi.m303chartdemo I/MPAndroidChart: Preparing...
08-01 16:08:53.361 20364-21189/com.qixi.m303chartdemo I/MPAndroidChart: offsetLeft: 122.0, offsetTop: 0.0, offsetRight: 0.0, offsetBottom: 146.0
08-01 16:08:53.361 20364-21189/com.qixi.m303chartdemo I/MPAndroidChart: Content: RectF(122.0, 52.5, -52.5, -146.0)
08-01 16:08:53.361 20364-21189/com.qixi.m303chartdemo I/MPAndroidChart: Preparing Value-Px Matrix, xmin: 1.4699814E12, xmax: 1.4700376E12, xdelta: 5.6229888E7
08-01 16:08:53.361 20364-21189/com.qixi.m303chartdemo I/MPAndroidChart: Data is set.
08-01 16:08:53.361 20364-21189/com.qixi.m303chartdemo D/MainActivity: co2 arrayCo2.size():64
08-01 16:08:53.361 20364-21189/com.qixi.m303chartdemo D/HttpUtil: disconnect
08-01 16:08:53.361 20364-20364/com.qixi.m303chartdemo E/AndroidRuntime: FATAL EXCEPTION: main
08-01 16:08:53.361 20364-20364/com.qixi.m303chartdemo E/AndroidRuntime: Process: com.qixi.m303chartdemo, PID: 20364
[...]

Issue #2070, opened 2016-07-26
IndexOutOfBoundsException when empty LineDataSet in LineData of LineChart


lschuetze:
Component: LineChart
Version: v3.0.0-beta1

The `LineChart` will throw an exception when it is filled with `LineData` containing a `LineDataSet` with no entries. As soon as an entry is added upon creation it will not throw the exception.

Is this intented behavior?

Working code:
```java
ArrayList<Entry> data = new ArrayList<>();
data.add(new Entry(0, 0.0f));
throughputDataSet = new LineDataSet(data, "throughput");
```

Exception throwing code:
```java
ArrayList<Entry> data = new ArrayList<>();
throughputDataSet = new LineDataSet(data, "throughput");
```

The exception trace:
```java
java.lang.IndexOutOfBoundsException: Invalid index 17, size is 17
 at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
 at java.util.ArrayList.get(ArrayList.java:308)
 at com.github.mikephil.charting.formatter.FormattedStringCache$PrimFloat.getFormattedValue(Unknown)
 at com.github.mikephil.charting.formatter.DefaultAxisValueFormatter.getFormattedValue(Unknown)
 at com.github.mikephil.charting.components.AxisBase.getFormattedLabel(Unknown)
 at com.github.mikephil.charting.components.AxisBase.getLongestLabel(Unknown)
 at com.github.mikephil.charting.renderer.XAxisRenderer.computeSize(Unknown)
 at com.github.mikephil.charting.renderer.XAxisRenderer.computeAxisValues(Unknown)
 at com.github.mikephil.charting.renderer.XAxisRenderer.computeAxis(Unknown)
 at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(Unknown)
 at android.view.View.draw(View.java:15231)
 at android.view.View.updateDisplayListIfDirty(View.java:14167)
 at android.view.View.getDisplayList(View.java:14189)
 at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3389)
 at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3368)
 at android.view.View.updateDisplayListIfDirty(View.java:14127)
 at android.view.View.getDisplayList(View.java:14189)
 at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3389)
[...]

PR #2094, opened 2016-08-03
Fix for IndexOutOfBoundException of FormattedStringCache

lschuetze:
This pull request contains of a fixed version of `FormattedStringCache` and a fix of the respective test class.
`IndexOutOfBoundException` with FormattedStringCache classes have been reported in issue #2083 and #2070. 

### Problem description
The original version hold two `ArrayList` to store a mapping from a value to its formatted String representation. Therefore, the values where iterated to see if a value has already be seen before. The index calculated in the iteration is then used to access the formatted string cache. This resulted sometimes in an `IndexOutOfBoundsException`.

### Solution
 The fix is to use a `HashMap` to store the mapping. This reduces the time to find a value from O(n) to O(1). Furthermore, there are no more two data structures that have to be synced. On top of that the claimed avoidance of auto-boxing is now really forced.

The solution did not change any API. Just internal data structures have been changed.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/FormattedStringCache.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/FormattedStringCache.java
@@ -6,59 +6,60 @@ import java.util.HashMap;
 
 /**
  * Created by Tony Patino on 6/29/16.
+ * <p>
+ * COST : Frequently the V type, and the K type will often be passed as primitives (float / int / double)
+ * This will incur an auto-boxing penalty, and an instantiation on each call.
+ * <p>
+ * BENEFIT : Formatting of Strings is one of the costliest operations remaining, and it is larger than the boxing penalty.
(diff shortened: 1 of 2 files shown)


tonypatino-monoclesociety:
@lschuetze As long as using Float or Double as keys in the HashMap instances won't result in issues, I think that your HashMap approach is preferable.

lschuetze:
Since HashMap uses hashCode this could indeed result in Float representing the same value are not regarded the same. However, in that case it would result in values being computed again. This happens when both Float values differ in a small epsilon. The performance would still be better regarding that an ever growing ArrayList results in Array copying due to its implementation.

An alternative would require a one to one comparison using an epsilon which is O(n) again.

lschuetze:
Since `Float` as HashMap key makes not so much sense and commit 02bf21df50f2d31f3f9c7276b24aba9a3dae163e I will close this pull request as it is not relevant anymore.

PR #2110, opened 2016-08-07
AxisLine DashEffect

farfromrefug:
I added method to allow dash effect on axis line, just like on iOS.
Also i added methods to set the ```DashPathEffect``` directly as i needed it.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/components/AxisBase.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/components/AxisBase.java
@@ -504,6 +504,17 @@ public abstract class AxisBase extends ComponentBase {
                 lineLength, spaceLength
         }, phase);
     }
+    
+    /**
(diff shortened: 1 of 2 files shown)


farfromrefug:
@PhilJay i have made a lot of modifications in my fork to make MPAndroidChart and Charts more alike. This is because i am building a module for the cross-platform framework Titanium Mobile
Can you take a quick look at the commits and tell me if that is something you want?
I am asking first because it is quite time consuming to create a branch?PR for every single commit...
Thanks

PhilJay:
Of course I am open to all kinds of improvements. What is important to me is that the change adds value to the library useful for a great amount of users. I cannot really incorporate changes that only effect a very small target group. 

Other than that, it is also important that proposed pull requests conform with the existing coding and documentation style of the library and are in proper form.

Nevertheless I cannot promise to merge any requests before I have reviewed them, but considering what I just said definitely makes it quite likely :-)

PR #2168, opened 2016-08-17
Fixed Bug in LineChartRenderer

Sin-Wave:
The LineChartRenderer didn't use x values to draw the cubic fill, but the x indices instead. This problem is fixed by my commit.


jqj1994:
There is still something wrong.  When animator starts but not arrive the max, the line will line to the max.
So your first line should change into
spline.lineTo(dataSet.getEntryForIndex(mXBounds.min**_+mXBounds.range_**).getX(), fillMin);

rather than 
spline.lineTo(dataSet.getEntryForIndex**_(mXBounds.max_**).getX(), fillMin);

Sin-Wave:
Now the animator should also work correctly.

PR #2193, opened 2016-08-24
Pie chart, fix for small values appearing as background

relair:
This fix is so small values won't appear as "background" (which really is a huge circle) when slice space is used.
I've found out that because sweepAngleOuter is negative it gets set to 0 which then causes arcStartPointX and arcStartPointY not to be set. Both arcStartPointX and arcStartPointY are used in calculateMinimumRadiusForSpacedSlice which calculates huge negative value because of that, which is then turned into huge positive value in the next line.
In the end we end up with a tiny value creating huge circle that looks like a background.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/PieChartRenderer.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/PieChartRenderer.java
@@ -263,15 +263,13 @@ public class PieChartRenderer extends DataRenderer {
 
                     mPathBuffer.reset();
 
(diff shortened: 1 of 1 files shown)


PhilJay:
Did you test this with different kind of data?
If yes, thank you, I will gladly merge.

relair:
I tested it for several cases different cases  - with separator, and no separator. I currently have it like that as an overridden method and it seems to work fine - we have it on production right now.
Essentially I found out the solution by comparing this library with its iOS mirror, as the issue didn't happen there.
I am not sure what other kinds of data could I test it with, could you please clarify?

Issue #2219, opened 2016-09-01
Maxint digits and Out Of Memory


Alebat:
Hi, I am working on an activity with a number of **real time LineCharts** with 100 entries each dynamically added and removed. 

An **OOM** happens, I debugged a bit and I found in a **DefaultAxisValueFormatter** a String Builder that tries to build a string with over 2 billions zeros. I found that this comes from the following lines of code when interval equals to 0.
```
        // set decimals
        if (interval < 1) {
            mAxis.mDecimals = (int) Math.ceil(-Math.log10(interval));
        } else {
            mAxis.mDecimals = 0;
        }
```
The log10 returns **-Infinity** the minus puts it to +Infinity and the int cast produces a **maxint** (2147483647).

I didn't manage to find a way to reproduce this problem but it happens on high data rates (3 to 6 charts @ from 50 to 250 Hz). It does not happen if the execution is slowed down by some conditional breakpoints. I put and remove the data from a non-ui thread. I call the notifyDataSetChanged, xrange and shift to last value every 33 millis in a Runnable posted to the UI thread.

I personally think that some debug may be done on the previous lines that work with the variable **interval** or just changing the code as follows: 
```
        // set decimals
        if (interval < 1 && interval > 0) {
            mAxis.mDecimals = (int) Math.ceil(-Math.log10(interval));
        } else {
            mAxis.mDecimals = 0;
        }
```
I am available to post details like a stack trace and so on.

Alebat:
I tried to use this code as a workaround to solve the problem, given that I don't actually need the X axis to be drawn:
```
        lineChart.getXAxis().setEnabled(false);
```
but the axis _seems to be computed in **any case**_ generating another **OOM** with a different stack trace, this time from user code:

**Previous error:**
```
java.lang.OutOfMemoryError: [...]
    at java.lang.AbstractStringBuilder.enlargeBuffer(AbstractStringBuilder.java:94)
    at java.lang.AbstractStringBuilder.append0(AbstractStringBuilder.java:145)
    at java.lang.StringBuffer.append(StringBuffer.java:219)
    at com.github.mikephil.charting.formatter.DefaultAxisValueFormatter.<init>(DefaultAxisValueFormatter.java:35)
    at .charting.components.AxisBase.getValueFormatter(AxisBase.java:482)
    at .charting.components.AxisBase.getFormattedLabel(AxisBase.java:451)
[...]

PR #2225, opened 2016-09-02
Issue #2219

Alebat:
#2219 

#### Avoids the axis to be updated when they are disabled (by Chart#notifyDatasetChanged)


Alebat:
Found problems

Issue #2261, opened 2016-09-14
LargeValueFormatter crashes if values exceed specified suffixes


cpboyd:
I'm working on a PR to fix this, but the issue is as stated:
If you use a number large enough to exceed the suffix array, then it causes a crash.

This seemed a bit surprising for a "LargeValue" formatter.

My PR will fallback to scientific notation in the event of such an overflow.

It might be a bit jarring to see a transition from "t" to "E14" but it seems better than simply crashing.

lloydtorres:
This issue still exists in the v3.0.0 release, but didn't exist in v2.2.5 and below. I think @cpboyd's PR is a good solution for dealing with overly large numbers.

PR #2262, opened 2016-09-14
LargeValueFormatter Fixes

cpboyd:
This PR should fix the crashes I reported in issue #2261.

This should eliminate any crashes with formatting very big numbers by falling back to scientific/exponent notation if there’s not an appropriate suffix.

Additionally, it should also better handle user suffixes that aren’t simply lowercase letters.


almic:
I know it's been a VERY long time but if you want, please update this to resolve the conflicts. I would do it myself but I'm trying to get this project back to a manageable state and I cannot spend too long on every pull request and issue.

PR #2439, opened 2016-11-04
Fixed NPE bug with device on SDK 10 #2259

dzungpv:
Fixed NPE bug with device on SDK 10 #2259


danielgindi:
What about code styling?

almic:
The project `minSdkVersion` has been increased to 14, as there are now less than 1% of devices running lower than version 14.

Issue #2455, opened 2016-11-10
LineChart | Crash if LineDataSet is empty.


hitech95:
Hi,
I just updated from beta to stable and now the activity crash.
After a few of tests I have found that append if a LineDataSet is empty and is added to a LineData.

```
public class ChartActivity extends AppCompatActivity {
    @BindView(R.id.chart)
    LineChart mChart;

    LineData mData;
    LineDataSet mDataSet;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_chart);

        ButterKnife.bind(this);

        mData = new LineData();
        mChart.setData(mData);
        mChart.invalidate();

        mDataSet = new LineDataSet(null, "Weight");

        // If you add a value works
        mDataSet.addEntry(new Entry(0, 5));

        mData.addDataSet(mDataSet);
        mData.notifyDataChanged();
    }
}
```
```
FATAL EXCEPTION: main
    Process: it.rblab.mydemochart, PID: 6312
    java.lang.IndexOutOfBoundsException: Invalid index 0, size is 0
     at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
     at java.util.ArrayList.get(ArrayList.java:308)
     at com.github.mikephil.charting.data.DataSet.getEntryForIndex(DataSet.java:286)
     at com.github.mikephil.charting.utils.Transformer.generateTransformedValuesLine(Transformer.java:184)
     at com.github.mikephil.charting.renderer.LineChartRenderer.drawValues(LineChartRenderer.java:545)
     at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(BarLineChartBase.java:251)
     at android.view.View.draw(View.java)
     at android.view.View.updateDisplayListIfDirty(View.java)
     at android.view.View.getDisplayList(View.java)
     at android.view.View.draw(View.java)
     at android.view.ViewGroup.drawChild(ViewGroup.java)
     at android.view.ViewGroup.dispatchDraw(ViewGroup.java)
     at android.view.View.updateDisplayListIfDirty(View.java)
     at android.view.View.getDisplayList(View.java)
     at android.view.View.draw(View.java)
     at android.view.ViewGroup.drawChild(ViewGroup.java)
     at android.view.ViewGroup.dispatchDraw(ViewGroup.java)
     at android.view.View.updateDisplayListIfDirty(View.java)
[...]

Issue #2450, opened 2016-11-09
Invalid index 0, size is 0 in lineChart


MaAnQing:
in the first, I add an empty chart data ,then

    java.lang.IndexOutOfBoundsException: Invalid index 0, size is 0
    11-09 19:34:42.397 E/AndroidRuntime( 5201): 	at             java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
    11-09 19:34:42.397 E/AndroidRuntime( 5201): 	at java.util.ArrayList.get(ArrayList.java:308)
    11-09 19:34:42.397 E/AndroidRuntime( 5201): 	at com.github.mikephil.charting.data.DataSet.getEntryForIndex(DataSet.java:286)
    11-09 19:34:42.397 E/AndroidRuntime( 5201): 	at com.github.mikephil.charting.utils.Transformer.generateTransformedValuesLine(Transformer.java:184)
    11-09 19:34:42.397 E/AndroidRuntime( 5201): 	at com.github.mikephil.charting.renderer.LineChartRenderer.drawValues(LineChartRenderer.java:545)
    11-09 19:34:42.397 E/AndroidRuntime( 5201): 	at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(BarLineChartBase.java:251)

I think it is because of you dont check size when you draw values

MaAnQing:
in 3.0.0 version

pedrovieira7:
I'm also getting this @PhilJay, any workarounds?

prometey123:
I'm using v.3.0.2 "compile 'com.github.PhilJay:MPAndroidChart:v3.0.2'"  but the bug is still present

fernando-s97:
It seems the problem is in version 3.0.2. After getting back to version 3.0.1, the problem got gone.

jvlichai:
Yeah, I happens on Pixel XL device API 25(Android 7.1) when use 3.0.2. After use version 3.0.1 no problem.

madrzazg:
I to have this problem with 3.0.2. Changing back to 3.0.1 does not fix anything, unfortunately.

TolikB:
I to have this problem with 3.0.2. Changing back to 3.0.1 does not fix anything, unfortunately.
--

i have same problem

animalecs:
Are there any workarounds for this problem?
I'm facing this bug even if I check that the list passed to the chart is not null and with at least one element.
[...]

PR #2462, opened 2016-11-12
Fix empty chart data handling

monz:
Hi Phil,

first of all thank you for your great library, really appreciate it!

Maybe I could solve two little issues, see commits. All tests ran successfully.
I had the same issues as in #2450 and #2455 which should work now.
Additional, if the dataset was empty but not null the "no data text" was not shown.

Greetings,
Markus


GoneUp:
+1 would be really usefull

Philipp2405:
+ 1

zhiwo:
+1 would be really usefull

PR #2507, opened 2016-11-24
Added LimitRectangle

ZmorzynskiK:
Added LimitRectangle - similar to LimitLine but instead of drawing lines, we are drawing rectangles (might be semi-transparent).


ZmorzynskiK:
Just wanted to say that I'm using `LimitRectangle` in one of my projects and I think it all looks good :)

PR #2596, opened 2016-12-16
Package ProGuard rules into the aar.

monxalo:


andhie:
This prevents the whole library from being Proguard.
Should use a stricter rule. Solution specified in https://github.com/PhilJay/MPAndroidChart/issues/348#issuecomment-87453245

```java
-keep public class com.github.mikephil.charting.animation.* {
    public protected *;
}
```

monxalo:
I just followed the proguard rules from the [wiki](https://github.com/PhilJay/MPAndroidChart/wiki/Proguard), which says that it needs to whitelist the library but maybe there are better rules.

hannesa2:
@monxalo and @andhie Thank you, for me it works.

```
allprojects {
	repositories {
		...
	}
}
```

and 

andhie:
you need to remove the original rule `-keep class com.github.mikephil.charting.** { *; }`
if you have this, it still means keeping everything even you have a stricter rule

hannesa2:
Thank you for the hint
https://github.com/hannesa2/MPAndroidChart/commit/115cdb9105e6e496976c0b778cb4bfaa469ba87f

almic:
This doesn't seem necessary with Android Studio, is this specific to Eclipse or some other thing? I deleted these files in a recent commit because it was causing the Android lint to give warnings.

sonique6784:
@monxalo can you please fix conflict?

MPChartLib/proguard-project.txt
```diff
@@ -1,20 +1,4 @@
...
 
-# Add any project specific keep options here:
-
-# If your project uses WebView with JS, uncomment the following
-# and specify the fully qualified class name to the JavaScript interface
-# class:
-#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
-#   public *;
-#}
+-keep class com.github.mikephil.charting.** { *; }
```
sonique6784: following rules should be used:
`-dontwarn com.github.mikephil.**`

`-keep public class com.github.mikephil.charting.animation.* {
    public protected *;
}`

PR #2667, opened 2017-01-05
Pass entry as argument in IShapeRenderer.renderShape

philippeauriach:
This allows users to show different shapes depending of the entry. For example, we could show the same shape but with a different gradient of a colour depending on the y value of the entry.


almic:
Could you please create a new pull request for this, as the file structure has changed. I think this is good, but if you could also make use of it in the `CustomScatterShapeRenderer` to show this gradient color you describe.

PR #2681, opened 2017-01-07
Support Yaxis set diffrent text color of up and down lable text

spykjl:
Support Yaxis set diffrent text color of up and down lable text,add use demo to LineChartActivity1


spykjl:
This allows users to show different color of Yaxis label color,my project need that function.I remember the old version is support that,but v3.0.1 is not support.

Issue #2243, opened 2016-09-08
[bug and fix] avoidFirstLastClippingEnabled not working properly


beautifulSoup:
In XAxisRenderer there is

```java
if (i == mXAxis.mEntryCount - 1 && mXAxis.mEntryCount > 1) {
                        float width = Utils.calcTextWidth(mAxisLabelPaint, label);

                        if (width > mViewPortHandler.offsetRight() * 2
                                && x + width > mViewPortHandler.getChartWidth())
                            x -= width / 2;

                        // avoid clipping of the first
                    } else if (i == 0) {
```

the judge can't get the last entry properly, because i is double of the entries count. So it should be 

```java
if (i / 2 == mXAxis.mEntryCount - 1 && mXAxis.mEntryCount > 1) {
                        float width = Utils.calcTextWidth(mAxisLabelPaint, label);

                        if (width > mViewPortHandler.offsetRight() * 2
                                && x + width > mViewPortHandler.getChartWidth())
                            x -= width / 2;

                        // avoid clipping of the first
                    } else if (i == 0) {
```

I think.

Zarwan:
Has this been resolved anywhere? I'm currently encountering this problem when trying to upgrade to the new release from 2.5. I have tried using `avoidFirstLastClippingEnabled` but it doesn't seem to do anything.

What used to happen:
![screen shot 2016-11-01 at 4 50 35 pm](https://cloud.githubusercontent.com/assets/8028687/19906972/5e1192a4-a053-11e6-9f81-9a677a71340a.png)

What is happening now:
![screen shot 2016-11-01 at 4 50 20 pm](https://cloud.githubusercontent.com/assets/8028687/19906977/632a73a0-a053-11e6-99d8-f4cd77465575.png)

DrewSz:
Thank you so much, I spent forever trying to figure this out. Dividing i by 2 solved the problem.

PR #2719, opened 2017-01-16
fix the bug that avoidFirstLastClippingEnabled not working properly

beautifulSoup:
For i is double to the entryCount, to get the last entry, the (mXAxis.mEntryCount - 1) should be also doubled. see #2243


stinsonga:
Any status updates on this particular issue? It still seems to be an outstanding issue in the lib.

almic:
@beautifulSoup Thanks for keeping up with the library, but sadly I already merged another pull request (#3820) that fixed this. I've been working from newest to oldest issues and PRs, so that's why I merged the other one before this.

Hope you understand 👍

PR #2799, opened 2017-02-11
About function"isDrawingValuesAllowed"

xiezhigang:
Hi,
Please ignore my poorly English.

I think it is better that replace the function "isDrawingValuesAllowed"  with my code below in situation which like the picture below.

`protected boolean isDrawingValuesAllowed(ChartInterface chart) { return chart.getXRange() < chart.getMaxVisibleCount() * mViewPortHandler.getScaleX(); /*return chart.getData().getEntryCount() < chart.getMaxVisibleCount() * mViewPortHandler.getScaleX();*/ }`

[picture](https://github.com/xiezhigang/Android/blob/master/issue.png)


almic:
I think the current function works the way it should.

Tell me if you think this should really be changed though, maybe added as an option.

Issue #441, opened 2015-03-16
Fill LineChart sometimes not working


warmax:
With last version 2.0.8 sometimes the chart is not filled. When I revert back to 2.0.7 it works again.

set1.setDrawFilled(true);
set1.setFillColor(getResources().getColor(R.color.red_down));
mChart.setData(data)
mChart.invalidate();

regards

PhilJay:
Could you describe the situation in more detail?
Which setup (which data) causes the chart not to be filled correctly?
Could you provide a screenshot of what it looks like?

warmax:
Data is coming from a financial api so it's changing quite a bit.... some charts are just fine but others are not filled (I haven't been able to find a pattern). I attach a screenshot with v2.0.7 and v2.0.8 on the same chart:

![mp207](https://cloud.githubusercontent.com/assets/9325874/6699637/3f3c3498-cd04-11e4-922b-53807417531b.png)

![mp_208](https://cloud.githubusercontent.com/assets/9325874/6699640/442e3c80-cd04-11e4-837f-0a36c3ec00ba.png)

If you have any ideas of how can I help, please let me know.

PhilJay:
Okay so the whole area is simply not filled. That is quite weird.
How many values are you plotting?

If the issue occurs, could you tell me if you can find the following warning message in your logcat: "Path too large to be rendered into a texture"

Thank you!

warmax:
Yes I'm getting that error in logcat.

491 values -> not working
515 values -> not working
1545 values -> not working
128 values ->working 
261 values ->working
781 values -> working, strange??

xu6148152:
may be you can debug the view which can not be working well using your data, and this way may solve this problem fast.

PhilJay:
Alright. Does the issue occur instantly, or only when zooming in?

PhilJay:
Could you try the latest commit and see if you are still having the issue?

warmax:
Hi phil, just checked with last commit and it worked fine! 

Thanks

PhilJay:
Good to know, I thought I might have fixed it!

gray419:
I am still having this exact issue, it was working in 2.0.7, just updated my gradle build to 2.0.8 today and got the irregular fill behavior:

These are two diff charts with fill working and not working:
[...]

PR #2837, opened 2017-02-21
#441 Fill not working

semirke:
This pull request is a proposed to solve #441  (https://github.com/PhilJay/MPAndroidChart/issues/441)
Basically the vanishing fillings were caused by scattered Float.NaNs. 
I tested this solution with NaNs missing from the beginning, end and in  between, however, to much and scattered NaNs could maybe still render a graph useless. 

Ps: sorry for the in between commits, my git is a bit rusty...

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/LineChartRenderer.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/LineChartRenderer.java
@@ -481,7 +481,8 @@ public class LineChartRenderer extends LineRadarRenderer {
      */
     private void generateFilledPath(final ILineDataSet dataSet, final int startIndex, final int endIndex, final Path outputPath) {
 
-        final float fillMin = dataSet.getFillFormatter().getFillLinePosition(dataSet, mChart);
+        // TODO just to check my skills with git+gradle :)
+        final float fillMin = 30;//dataSet.getFillFormatter().getFillLinePosition(dataSet, mChart);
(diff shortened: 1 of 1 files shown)


MPChartLib/src/main/java/com/github/mikephil/charting/renderer/LineChartRenderer.java
```diff
@@ -488,18 +488,24 @@ private void generateFilledPath(final ILineDataSet dataSet, final int startIndex
         final Path filled = outputPath;
         filled.reset();
 
-        final Entry entry = dataSet.getEntryForIndex(startIndex);
+        // Semirke: we must skip Float.NaNs first
+        int sIdx = startIndex;
+        while(Float.isNaN(dataSet.getEntryForIndex(sIdx).getY())
+            && sIdx < endIndex ) sIdx++;
+
```
semirke: I could've added a check for success.

noberasco:
Hi @semirke,

I'm trying your fix as I'm affected by the same issue.
However, I find that the LineChart issue with NaN values is not limited to the chart fill.
If the portion of LineChart visible on screen starts with a NaN, than the whole chart won't be visible until you move/zoom it so that it starts with a valid number.

noberasco:
Here is a quick fix for my chart-not-drawing issue:
#2878

almic:
Thank you for this, but there's been changes to the repo since this. If you resolve the conflicts, then I can merge it.

PR #2842, opened 2017-02-22
Implementation of equals() and hashcode() for class Entry

PaddySmalls:
Hi,
I recently ran into into an issue when I implemented some unit tests for our university project ["Zeitfresser"](https://github.com/slipke/Zeitfresser). I had to LinkedLists containing two Entry instances with the exact same property values. However, comparing these two Lists always have me _false_ as a return value, since every call to _equals()_ on class _Entry_ is delegated to _Object_.
I thought that implementing these methods for _Entry_ and its base class _BaseEntry_ might be valuable not just for my own use case. Having and comparing Lists of Entries seems like a common scenario from my perspective. 
For the implementation of both methods, I followed common standards and best practices.
In oder to verify _equals()_ as well as _hashcode()_ work properly, I also added a test class _EntryTest_.  
This class covers the most basic cases for Entry comparison. Of course I also made sure that my changes don't break any other running tests.  

Cheers, Patrick

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/data/BaseEntry.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/data/BaseEntry.java
@@ -129,11 +129,19 @@ public abstract class BaseEntry {
         return false;
       }
 
-      if ((other.getData() != null && getData() == null) && !other.getData().equals(getData())) {
+      if (getData() != null && !getData().equals(other.getData())) {
         return false;
       }
 
-      if ((other.getIcon() != null && getIcon() != null) && !other.getIcon().equals(getIcon())) {
+      if (other.getData() != null && !other.getData().equals(getData())) {
+        return false;
+      }
+
+      if (getIcon() != null && !getIcon().equals(other.getIcon())) {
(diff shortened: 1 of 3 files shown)


Alex-Michie:
It may be worth noting that this implementation of hashcode is based on mutable fields, making it unsuitable for general use.

while if your aware that this hashcode function can change depending on what you do with the entry class it wont matter, this differs from the function your overriding and therefore may not be a suitable implementation.

with that said, i dont see a good way to implement the hashcode function because there are no immutable fields here aside from where the object is allocated (the default hashcode), so i dont think there is a way to implement hashcode and equals without introducing side effects. perhaps creating a new function which does the equality check which you needed in this case would be the way to go?

PaddySmalls:
Agreed, we worked around that problem anyway. The hashcode func didn't bother us in our scenario, since we just built a LinkedList of entries and displayed them in the Activity. For general purpose, I should have double-thought my approach.
However, introducing something like a class-wide identifier for scenarios where a certain entry must be referenced without having the instance's reference available could make sense in my opinion.

PR #2855, opened 2017-02-28
We need only LegendEntry base Label and formColor

jhamin0511:


Rainer-Lang:
@PhilJay Could this be merged?

almic:
There are a lot of unnecessary additions to the javadocs here, and I don't think this is the best way to go about this. If you submit a new pull request with only the additional constructor, then I'll consider it.

PR #2856, opened 2017-02-28
Fix for missing setters in getInstance method

stephenmcbride:
The zoomAndCenterAnimated method in BarLineChartBase crashes with a NullPointer exception because the yAxis variable is null when onAnimationUpdate is called. The yAxis is null because of missing setters in the getInstance method of AnimatedZoomJob.


idealllee:
hitting the same issue. Fix looks good (Thanks @stephenmcbride). @PhilJay plans to merge it anytime soon?

dpeters1:
Thanks @stephenmcbride! Would love to see this merged to master.

almic:
Looks right to me. Merging.

PR #2866, opened 2017-03-01
Use Property classes instead of reflection to handle animations

plackemacher:
This should remove any necessity to declare ProGuard rules for this library as this removes the reflection lookup needed when animating the four different properties here that are currently specified by strings. The downside to this change is that the minimum SDK required for the library would need to change to 14.

I'm putting this out there in case changing the minimum SDK to 14 is acceptable.

### DIFF
--- a/MPChartLib/build.gradle
+++ b/MPChartLib/build.gradle
@@ -8,7 +8,7 @@ android {
     buildToolsVersion '23.0.3'
     // resourcePrefix 'mpcht'
     defaultConfig {
-        minSdkVersion 9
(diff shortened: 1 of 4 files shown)


MGaetan89:
If `minSdk` is now `14`, there are some code that can be simplified.
At least here: https://github.com/PhilJay/MPAndroidChart/search?q=SDK_INT&type=Code
But I guess there may be other places as well.

almic:
There has been changes to the repo since this pull request, if you could resolve the conflicts then I can merge this. Thank you!

PR #2878, opened 2017-03-03
Fix for chart not being drawn if visible portion starts with a NaN

noberasco:


JKorsten:
Is this also possible for the barchart? Now the when I have NaN values my data isn't positioned correctly

noberasco:
resynced with current master, plus added additional fix for chart fill

JKorsten:
Why isn't this in the current release yet?

almic:
@noberasco the pull request by semirke already does this

biosignalssolutions:
I can't plot NaN value for LineChart on android API29. (samsung A41)   using set.getEntryForIndex(i).setY(NaN);
With android API23 (BQ Aquaris E5) works without any change at code.

Any solution/comment, please.
Thanks.

PR #2921, opened 2017-03-15
Fixed issue with filled and stepped mode. 

JacobJust:
Fixed issue with filled and stepped mode. 
The first one in each iteration would be linear instead of filled.


JacobJust:
device-2017-03-15-124217 before the fix

JacobJust:
device-2017-03-15-141741
after the fix

tatocaster:
This is not shipped yet. and the bug still exists. @PhilJay

PR #2946, opened 2017-03-21
Make LargeValueFormatter Locale aware / fixed corresponding Test 

Iridias:
In case that default Locale is not English but e.g. German, the LargeValueFormatterTest would fail due to expecting the result "5.5" but getting "5,5"!
This prevented building the project for german developers (and maybe in some other countries too).


sunnykinger:
In Kazakhstan, they don't use b or m for billion and million and instead they use Млн for million and Млрд for billion. Can you suggest any good way on how I can achieve this properly?

Issue #2952, opened 2017-03-22
IndexOutOfBoundsException


MaAnQing:
``` java
java.lang.IndexOutOfBoundsException: Invalid index 0, size is 0
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at java.util.ArrayList.get(ArrayList.java:308)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at com.github.mikephil.charting.data.DataSet.getEntryForIndex(DataSet.java:286)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at com.github.mikephil.charting.utils.Transformer.generateTransformedValuesLine(Transformer.java:184)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at com.github.mikephil.charting.renderer.LineChartRenderer.drawValues(LineChartRenderer.java:545)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(BarLineChartBase.java:251)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.View.draw(View.java:16224)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.View.updateDisplayListIfDirty(View.java:15221)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.View.draw(View.java:15994)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.ViewGroup.drawChild(ViewGroup.java:3611)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3401)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.View.draw(View.java:16227)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.View.updateDisplayListIfDirty(View.java:15221)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.View.draw(View.java:15994)
03-22 21:34:37.714 E/AndroidRuntime( 6445): 	at android.view.ViewGroup.drawChild(ViewGroup.java:3611)
[...]

PR #2958, opened 2017-03-26
fix issues 2952 : IndexOutOfBoundsException

MaAnQing:
fixed issues https://github.com/PhilJay/MPAndroidChart/issues/2952

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/data/DataSet.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/data/DataSet.java
@@ -283,6 +283,9 @@ public abstract class DataSet<T extends Entry> extends BaseDataSet<T> {
 
     @Override
     public T getEntryForIndex(int index) {
+        if (mValues == null || mValues.isEmpty()) {
+            return null;
+        }
         return mValues.get(index);
     }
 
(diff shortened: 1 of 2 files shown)


renannery:
Not sure if this PR will fixed the issue, once returns null, the hierarchy of calls will crash as well expecting a not null value.

anyway, I'm not a fan of reflections, but here is the way I "fixed":

```
    @Override
    public T getEntryForIndex(int index) {
        if (mValues == null || mValues.isEmpty()) {
            return (T) getEmptyEntry();
        }

        return mValues.get(index);
    }

    private Entry getEmptyEntry() {
        Type type = ((ParameterizedType) getClass().getGenericSuperclass()).getActualTypeArguments()[0];

        if (type == BarEntry.class) {
            return new BarEntry(0, 0);
        }

        if (type == BubbleEntry.class) {
            return new BubbleEntry(0, 0, 0);
        }

        if (type == CandleEntry.class) {
            return new CandleEntry(0, 0, 0, 0, 0);
        }

        if (type == PieEntry.class) {
            return new PieEntry(0);
        }

        if (type == RadarEntry.class) {
            return new RadarEntry(0);
        }

        return new Entry(0, 0);
    }
```
Let me know if it makes sense and I can create a PR or modified the current one

MGaetan89:
@PhilJay Any chance this get reviewed/fixed for the next release? I'm also getting crash reports for this  issue.

Jawnnypoo:
@renannery I have had to pull this change in manually due to tons of crashing on an app, and it works just fine as is. @PhilJay This should really be pulled in with a 3.0.3 release. Its a pretty big bug with little chance of being fixed on the app side, would really appreciate it so that I can remove the forked code.

Issue #2917, opened 2017-03-15
  java.lang.NullPointerException: Attempt to invoke interface method 'int com.github.mikephil.charting.interfaces.datasets.IDataSet.getEntryIndex(com.github.mikephil.charting.data.Entry)


neocameback:
Hi @PhilJay I got this error when using Line chart.
Could you or someone check out for the problem. 

I think the problem is in these lines of your Chart class

`IDataSet set = mData.getDataSetByIndex(highlight.getDataSetIndex());

   Entry e = mData.getEntryForHighlight(mIndicesToHighlight[i]);
   int entryIndex = set.getEntryIndex(e);`

Where we can't get data set from input data.

Here is my code.

`Process: com.gemvietnam.canhbaochitieu, PID: 25781
                                                                               java.lang.NullPointerException: Attempt to invoke interface method 'int com.github.mikephil.charting.interfaces.datasets.IDataSet.getEntryIndex(com.github.mikephil.charting.data.Entry)' on a null object reference
                                                                                   at com.github.mikephil.charting.charts.Chart.drawMarkers(Chart.java:731)
                                                                                   at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(BarLineChartBase.java:268)
                                                                                   at android.view.View.draw(View.java:16201)
                                                                                   at android.view.View.updateDisplayListIfDirty(View.java:15198)
                                                                                   at android.view.View.draw(View.java:15971)`

`mLineChart.resetTracking();

//    int max = 1;
    ArrayList<ILineDataSet> dataSets = new ArrayList<>();
    List<THLKLineChartLegend> chartLegends = new ArrayList<>();
    List<String> dates = new ArrayList<>();

    // Build x Axis values
    Map<String, Integer> xAxisValues = Charts.getDateMapLine(accDivPlanChart.getLines(), dates);

    if (accDivPlanChart.getLines() != null && !accDivPlanChart.getLines().isEmpty()) {
      for (int z = 0; z < accDivPlanChart.getLines().size(); z++) {
        // Data for 1 line
        LChartLine chartLine = accDivPlanChart.getLines().get(z);
        ArrayList<Entry> values = new ArrayList<>();

        // Legend info
        String infoString = "";
        if (chartLine.getInfoList() != null) {

          for (LChartLineInfo info : chartLine.getInfoList()) {
            infoString += info.getKey() + " " + info.getValue() + ",";
          }

[...]

PR #3145, opened 2017-05-26
Fix null pointer exception in draw markers

JacobJust:
Fixed this issue: https://github.com/PhilJay/MPAndroidChart/issues/2917

We get crashes from our customers, but we are not able to reproduce the issues. Added a null check, for the variable set, which can be null.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/charts/Chart.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/charts/Chart.java
@@ -726,6 +726,8 @@ public abstract class Chart<T extends ChartData<? extends IDataSet<? extends Ent
             Highlight highlight = mIndicesToHighlight[i];
 
             IDataSet set = mData.getDataSetByIndex(highlight.getDataSetIndex());
+			if (set == null) 
+				continue;
 
             Entry e = mData.getEntryForHighlight(mIndicesToHighlight[i]);
             int entryIndex = set.getEntryIndex(e);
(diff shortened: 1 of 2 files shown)


cubizh:
This seems to be a nobrainer PR.
This issue happens to me when I remove the line where the marker is on at the moment.
I managed to get around this bug by .setDrawMarkers(false); on chart creation and only doing .setDrawMarkers(true); inside onValueSelected()

RaviUmraliya:
"This seems to be a nobrainer PR.
This issue happens to me when I remove the line where the marker is on at the moment.
I managed to get around this bug by .setDrawMarkers(false); on chart creation and only doing .setDrawMarkers(true); inside onValueSelected()"

@cubizh  ya it's work

alejouribesanchez:
> This seems to be a nobrainer PR.
> This issue happens to me when I remove the line where the marker is on at the moment.
> I managed to get around this bug by .setDrawMarkers(false); on chart creation and only doing .setDrawMarkers(true); inside onValueSelected()

This also works for me

Poojamalik810:
I am using MP chart library for behaving it like the real time graph to draw line graph in a real time way.


But the above solutions not working for me.  Infact, after setting .setDrawMarkers(false); on chart creation and only doing .setDrawMarkers(true); inside onValueSelected()
 I am not able to get any value when chart any point is clicked. That is, my marker donot give any value when clicked and it is crashing throwing this error.

I am continuously getting crash and getting this error.

**Kindly help as soon as possible how to handle this error**.

If anyone has used real time graph in android application. Kindly help me out.

Poojamalik810:
Help guys facing this issue from past many days as soon as possible

PR #3190, opened 2017-06-16
Axis multiline labels

fabriziogueli:
This PR allow multiline labels on the xAxis using '\n'

Example :

      XAxis xAxis = mChart.getXAxis();
      xAxis.setMultiLineLabel(true);

      xAxis.setValueFormatter(new IAxisValueFormatter() {
                @Override
               public String getFormattedValue(float value, AxisBase axis) {
                    return value+"\n"+"%";
                }
            });

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/components/XAxis.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/components/XAxis.java
@@ -47,6 +47,9 @@ public class XAxis extends AxisBase {
      */
     private boolean mAvoidFirstLastClipping = false;
 
+
+    private boolean mMultiLineLabel = false;
+
     /**
      * the position of the x-labels relative to the chart
      */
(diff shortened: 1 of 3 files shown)


ShamilovSV:
Maybe it will be better to use StaticLayout for multiline text instead of splitting by '\n'?

ClemMahe:
This modification would be useful. hope @PhilJay  or someone else will have a look :)

dhagz:
The texts are not center aligned.
Here are my updates to your PR.
In `Utils.java#drawXMultiLineText()`
* Change `paint.setTextAlign(Paint.Align.LEFT);` to `paint.setTextAlign(Paint.Align.CENTER);`
* Change `drawOffsetX += x;` to `drawOffsetX += x + (mDrawTextRectBuffer.width() / 2.0f);`

felislynx-silae:
It is very important feature! I hope someone will look at this. If you have problem with alignment, use static layout. From my experience it is much easier than calculating manually :)

joyceHong0524:
How can I edit this file if the Util.java is read-only file?
I need helppp :(

vicmpen:
@joyceHong0524 You need to clone, edit and build the project yourself

eunyeong1328:
I also tried setMultiLineLabel(true) but it doesn't work. For example?

saldous:
+1 interested in this. Please add!

PR #3232, opened 2017-07-04
Fix for crash we were getting when using MPAndroidChart library

PawelErnestLuczak:
Hey, this may be very little but from the crash reports we were getting there was a couple of crashes in this method. Usually array out of bounds exception


MGaetan89:
This is already proposed in #2958
But I think that a combination of the two would be best: check if `mValues` is null and if `index` is valid.

hannesa2:
@PawelErnestLuczak Thank you, for me it works.

```
allprojects {
	repositories {
		...
	}
}
```

and 

Issue #3259, opened 2017-07-15
how to prevent highlight be unselect by tap?


njleonzhang:
I don't want highlight to be removed by user's tap. How can I prevent it?

PR #3260, opened 2017-07-15
prevent deselect hightlighting per tap

njleonzhang:
#3259 add a flag `mHighLightDeselectPerTapEnable` to support


y2kappa:
Thanks for this.

drissfoo (changes requested):
No need for updating the gradle version and gradle.properties.
Everything else is perfect 👍🏽

fajardarwis (approved):

PR #3317, opened 2017-08-08
Added new coloring possibilities for LineChart.

xAqweRx:
There are possibly more chartes where this functionality could be added. But currently is needed only in LineChart
3 modes : default, horizontal gradient, vertical gradient

### DIFF
--- a/MPChartExample/res/values/color.xml
+++ b/MPChartExample/res/values/color.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="rating_store_9">#00ff00</color>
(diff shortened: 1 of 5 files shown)


xAqweRx:
As description what this new functionality do is : 
device-2017-08-08-095018

xAqweRx:
Real using looks like this 
device-2017-08-09-124059

reteps:
Looks good!!

sasisachin12:
> There are possibly more chartes where this functionality could be added. But currently is needed only in LineChart 3 modes : default, horizontal gradient, vertical gradient

@xAqweRx when it will be merge? we needs this changes.

PR #3384, opened 2017-08-28
Fix an Activity leak

scottkennedy:
If a ViewPortJob returns to the pool while holding a reference to a View, it will leak the Activity.

When we're recycling the jobs, we should just clear out all the object references.


scottkennedy:
Would it be possible to get this merged?

hannesa2:
@scottkennedy Thank you, for me it works.

```
allprojects {
	repositories {
		...
	}
}
```

and 

scott-kennedy-ck:
@almic would you be able to merge this?

SubhrajyotiSen:
Merged this at https://github.com/SubhrajyotiSen/MPAndroidChart which will be an updated fork

PR #3642, opened 2017-12-10
Option to set restrictions for Y axis autoscaling.

pagrzybe:
I need to be able to set some restriction on autoscale so I keep autoscale working but also make sure that certain range is always visible.

Tried to do it in the way the rest of the API is done and with as limited changes as possible.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
@@ -39,6 +39,27 @@ public class YAxis extends AxisBase {
     protected boolean mDrawZeroLine = false;
 
     /**
+     * flag indicating that auto scale min restriction should be used
+     */
+
+    private boolean mUseAutoScaleRestrictionMin = false;
+    /**
+     * flag indicating that auto scale max restriction should be used
+     */
+
+    private boolean mUseAutoScaleRestrictionMax = false;
+    /**
+     * restriction value of autoscale min
+     */
+
+    private float mAutoScaleMinRestriction = 0f;
+
+    /**
+     * restriction value of autoscale max
+     */
+    private float mAutoScaleMaxRestriction = 0f;
+
+    /**
      * Color of the zero line
      */
     protected int mZeroLineColor = Color.GRAY;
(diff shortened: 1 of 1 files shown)


MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -38,6 +38,27 @@
      */
     protected boolean mDrawZeroLine = false;
 
+    /**
+     * flag indicating that auto scale min restriction should be used
+     */
+
+    private boolean mUseAutoScaleRestrictionMin = false;
```
almic: set this to a single "use auto scale restriction" or something similar boolean flag.

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -38,6 +38,27 @@
...
+    /**
+     * flag indicating that auto scale max restriction should be used
+     */
+
+    private boolean mUseAutoScaleRestrictionMax = false;
```
almic: remove this one

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -38,6 +38,27 @@
...
+    /**
+     * restriction value of autoscale min
+     */
+
+    private float mAutoScaleMinRestriction = 0f;
```
almic: remove both of these two, `mAutoScaleMinRestriction` and `mAutoScaleMaxRestriction`

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -357,12 +378,54 @@ public boolean needsOffset() {
             return false;
     }
 
+    /**
+     * Sets min value restriction for autoscale
+     */
+    public void setAutoScaleMinRestriction(float restrictionValue) {
```
almic: remove both `setAutoScaleMinRestriction` and `setAutoScaleMaxRestriction` setters.

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -357,12 +378,54 @@ public boolean needsOffset() {
...
+     */
+    public void setAutoScaleMaxRestriction(float restrictionValue) {
+        mUseAutoScaleRestrictionMax = true;
+        mAutoScaleMaxRestriction = restrictionValue;
+    }
+
+    /**
+     * Resets min value restriction for autoscale
+     */
+    public void resetAutoScaleMinRestriction() {
```
almic: replace both `resetAutoScaleMinRestriction()` and `resetAutoScaleMaxRestriction()` with appropriate `setUseAutoScale???Restriction()` and `isUseAutoScale???Restriction()` getters and setters.

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
almic: remove both `mAutoScaleMinRestriction` and `mAutoScaleMaxRestriction` variables

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
almic: remove these two setters: `setAutoScaleMinRestriction()` and `setAutoScaleMaxRestriction()`

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -357,12 +378,54 @@ public boolean needsOffset() {
...
 
+        float min = dataMin;
+        float max = dataMax;
+
         // if custom, use value as is, else use data value
-        float min = mCustomAxisMin ? mAxisMinimum : dataMin;
-        float max = mCustomAxisMax ? mAxisMaximum : dataMax;
+        if( mCustomAxisMin ) {
+            min = mAxisMinimum;
+        } else if( mUseAutoScaleRestrictionMin ) {
```
almic: Indent this, so that the code checks for `mUseAutoScale???Restriction` ONLY if `mCustomAxisMin` is true, that way it uses the same `mAxisMinum` variable, but it two different possible ways.

MPChartLib/src/main/java/com/github/mikephil/charting/components/YAxis.java
```diff
@@ -357,12 +378,54 @@ public boolean needsOffset() {
...
+            min = Math.min( min, mAutoScaleMinRestriction );
+        }
+
+        if( mCustomAxisMax ) {
+            max = mAxisMaximum;
+        } else if( mUseAutoScaleRestrictionMax ) {
```
almic: Do the same with this as above.

pagrzybe:
Done and tested.

almic:
LGTM

almic:
Update for @pagrzybe, looks like this already existed with the methods `setAxisMaximum()` and `setAxisMinimum()`. I noticed a bug that causes graphs to gradually increase their y axis min and max range, mostly noticeable when dragging left-and-right on the seekbars to change the vertical range. This didn't exist on the horizontal range, however.

I think that the current implementation is good, but if you would like to I encourage you to see if the current version gives you what you wanted. If not, please open another pull request and test your changes by dragging the seekbars on the example app to make sure the ranges aren't drifting.

Sorry, and thanks for your patience!

PR #3676, opened 2017-12-20
add method to release static pool inside of MoveViewJob to avoid memory leak

svakulenko:
Hi 

This fix is for avoid memory leak by release of static ObjectPool instance ( and it array ) on activity onDestroy().
 
I attach chain of memory leak that happens in my activity.
With current fix, I have no more memory leaks.

mem_leak_stack

Best,


almic:
Is this actually a memory leak though? The class reserves two objects, which are reused as needed. Seems intentional that the objects stay in memory, because at any moment the user could do something that triggers an animation.

larten:
+1
I get this leak continuously from LeakCanary.
Will you merge this fix or is there any other solution?

PR #3603, opened 2017-11-22
Fix NPEs with WeakReference usage in PieChart

justindannguyen:
The issue happens because PieChart is added into scroll view of any kind of lazy init view.
When the view is lazy loading, its width and height is not available yet.
and NPE at [mDrawBitmap](https://github.com/PhilJay/MPAndroidChart/blob/master/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/PieChartRenderer.java#L638) which can only be created in [drawData](https://github.com/PhilJay/MPAndroidChart/blob/master/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/PieChartRenderer.java#L133) when ```width > 0 && height > 0```.

And got same problem and It works in my app.
This should fixes #3470

Issue #3470, opened 2017-09-30
Bug: PieChartRenderer - NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null object reference


Johnson145:
`mDrawBitmap` seems to be null and thereby causing a NullPointerException. Here is the full stacktrace:

`Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null object reference
       at com.github.mikephil.charting.renderer.PieChartRenderer.drawExtras(PieChartRenderer.java:638)
       at com.github.mikephil.charting.charts.PieChart.onDraw(PieChart.java:131)
       at android.view.View.draw(View.java:16238)
       at android.view.View.updateDisplayListIfDirty(View.java:15235)
       at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3637)
       at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3617)
       at android.view.View.updateDisplayListIfDirty(View.java:15195)
       at android.view.View.draw(View.java:16008)
       at android.view.ViewGroup.drawChild(ViewGroup.java:3653)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3443)
       at android.view.View.updateDisplayListIfDirty(View.java:15230)
       at android.view.View.draw(View.java:16008)
       at android.view.ViewGroup.drawChild(ViewGroup.java:3653)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3443)
       at android.view.View.updateDisplayListIfDirty(View.java:15230)
       at android.view.View.draw(View.java:16008)
       at android.view.ViewGroup.drawChild(ViewGroup.java:3653)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3443)
       at android.view.View.updateDisplayListIfDirty(View.java:15230)
       at android.view.View.draw(View.java:16008)
       at android.view.ViewGroup.drawChild(ViewGroup.java:3653)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3443)
       at android.view.View.updateDisplayListIfDirty(View.java:15230)
       at android.view.View.draw(View.java:16008)
       at android.view.ViewGroup.drawChild(ViewGroup.java:3653)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3443)
[...]

PR #3207, opened 2017-06-26
Fix some potential NPEs with WeakReference usage

scottkennedy:
Even if the WeakReference field is not null, the contained value may be null.  Additionally, you always need a strong reference to the value to ensure it isn't garbage collected while you're using it.

This should fix #3007.

PR #3763, opened 2018-01-29
Fix piechart NPE

Johnson145:
This is a bugfix for #3470. It's based on #3207 and an alternative solution to #3603. Instead of just cancelling the current request, it will try to fix the missing object references. Furthermore, some similar but not identical errors are solved, too.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/PieChartRenderer.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/PieChartRenderer.java
@@ -123,31 +123,18 @@ public class PieChartRenderer extends DataRenderer {
     @Override
     public void drawData(Canvas c) {
 
-        int width = (int) mViewPortHandler.getChartWidth();
-        int height = (int) mViewPortHandler.getChartHeight();
+        Bitmap drawBitmap = getStrongDrawBitmap();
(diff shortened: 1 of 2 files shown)


msapps:
I am getting crash mentioned in [#3825](https://github.com/PhilJay/MPAndroidChart/issues/3825) for line chart. 
I checked your commits and i noticed that you have done some changes in LineChartRenderer as well. So i was wondering if those changes can be solution to my problem?
Can you help me in this one?

Johnson145:
Actually, the changes in the LineChartRenderer have already been done in #3207. I just generalized that for the PieChartRenderer. However, to answer your actual question: Yes, I think you're problem can be solved by applying the same changes at the line(s) you're crash is occuring. I don't got the time to fix that one right now, but it should be straightforward for you to write a quick fix, if you have a look at the current changes. If you got any specific questions about it, feel free to ask. 

For the long run, it's probably the best solution to outsource the `getStrongDrawBitmap` method I wrote to a parent class or interface with default implementation to prevent having to handle this problem in each of the subclasses separately.

msapps:
Ok.. will do the changes and let you know in case of any queries.. Thanks a lot for help mate!

almic:
Merged 3207, closing.

PR #3807, opened 2018-02-14
Add ability to change formatter in LargeValueFormatter

ammaratef45:
I was testing my app when the default locale is Arabic, LargeValueFormatter returned me a non-meaning string, I wrote my own format class that default the DecimalFromat (mFormat) to use English instead of Arabic.
I thought it would be great if LargeValueFormatter class allows me to change the DecimalFormat object so I added it wishing that would be useful for other developers using this great library.
If my changes were useful I would be happy, So please let me know if that was a good thing to do or not.

Thanks
Ammar Atef

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/LargeValueFormatter.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/LargeValueFormatter.java
@@ -71,6 +71,16 @@ public class LargeValueFormatter implements IValueFormatter, IAxisValueFormatter
     public void setSuffix(String[] suff) {
         SUFFIX = suff;
     }
+	
+	/**
+     * Set custom DecimalFormat for formatting number.
+     * Default mFormat: DecimalFormat("###E00")
+     *
+     * @param format the new DecimalFormat
+     */
+    public void setDecimalFormatter(DecimalFormat format) {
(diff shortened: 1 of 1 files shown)


almic:
LargeValueFormatter is a convenience class to quickly format "large numbers in a pretty way." The idea is to create your own custom formatter by implementing `IValueFormatter` and `IAxisValueFormatter`.

If you have a value formatter that you want to add to the project, create a pull request for that.

almic:
You're looking for the file `strings.xml`. Do some reading about string localization at the [android developer website](https://developer.android.com/guide/topics/resources/localization#managing-strings).

You'll also want to look at [plural strings](https://developer.android.com/guide/topics/resources/string-resource#Plurals).

It's a little bit of reading, but if you want to implement localized functionality, you'll need to convert the current hard-coded strings in the `LargeValueFormatter` here:

```java
private static String[] SUFFIX = new String[]{
        "", "k", "m", "b", "t"
};
```

into string resources.

Please know this: I'm not sure if the `strings.xml` carries over into projects that declare this as a dependency, but this is definitely the way to go. I would start by adding this into your *own* project first, and then when you've got it working, try plugging that into your own fork of MPAndroidChart, make a release on GitHub, and see if the module handles it for you.

Best of luck!

PR #3809, opened 2018-02-15
optionally render stacked bar chart values with text color that match…

### LINKED ISSUE #2460: How to set text color of bar value as bar color for Stacked BarChart in MPChart
I'm using MP Chart library for my app. I have added stacked bar chart and value of all bar's are displayed in same color. I want to display the bar value in the bar color itself. Please help me how to achieve that.

image

mtotschnig:
…es data color

fixes #2460


ComputerKaige:
Please review this commit and merge to fix the value text colors as soon as possible, we need this.

mtotschnig:
@almic I have taken into account all your suggestions, with the exception of "And the boolean variable inside the new method can just be called enabled, not the name of the function.", which I do not understand.

Issue #3819, opened 2018-02-27
The setAvoidFirstLastClipping of XAxis doesn't work as expected


longup:
Hi:

I have issue on my phone, when I have setAvoidFirstLastClipping as true, the XAxis's last label entry
in the chart "clip" off the edge of the screen.

Here is my code:

lineChart.getXAxis().setAvoidFirstLastClipping(true);
        lineChart.getXAxis().setLabelCount(3,true);
        lineChart.setData(new LineData(dataSets));

        xDataList.add("2018-01.01");
        xDataList.add("2018-01.02");
        xDataList.add("2018-01.03");
        xDataList.add("2018-01.04");
        xDataList.add("2018-01.05");
        xDataList.add("2018-01.06");
        xDataList.add("2018-01.07");
        xDataList.add("2018-01.08");
        xDataList.add("2018-01.09");
        xDataList.add("2018-01.10");
        xDataList.add("2018-01.11");
        xDataList.add("2018-01.12");
        xDataList.add("2018-01.13");
        xDataList.add("2018-01.14");
        xDataList.add("2018-01.15");
        xDataList.add("2018-01.16");
        xDataList.add("2018-01.17");
        xDataList.add("2018-01.18");
        xDataList.add("2018-01.19");
        xDataList.add("2018-01.20");
        xDataList.add("2018-01.21");
        xDataList.add("2018-01.22");
        xDataList.add("2018-01.23");
        xDataList.add("2018-01.24");
        xDataList.add("2018-01.25");
        xDataList.add("2018-01.26");
        xDataList.add("2018-01.27");
        xDataList.add("2018-01.28");
        xDataList.add("2018-01.29");
        xDataList.add("2018-01.30");

        IAxisValueFormatter xFormatter = new IAxisValueFormatter() {
            @Override
            public String getFormattedValue(float value, AxisBase axis) {
                return xDataList.get((int)value);
            }
        };
        lineChart.getXAxis().setValueFormatter(xFormatter);

        final DecimalFormat df1 = new DecimalFormat("#.00");;

        IAxisValueFormatter yFormatter = new IAxisValueFormatter() {
            @Override
            public String getFormattedValue(float value, AxisBase axis) {
                return df1.format(value);
            }
        };
        lineChart.getAxisLeft().setValueFormatter(yFormatter);

I found that the method “drawLabels” in class XAxisRenderer has one bug, and I have fixed this issue. If you have the same issue, please update your library after I commit.

[...]

PR #3820, opened 2018-02-27
Avoid that the last label entry in the x-labels "clip" off the edge of the screen #3819

longup:
Modified the “drawLabels” methon in class XAxisRenderer, i should be i/2 in line 210, so the last label can be handled correctly.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/XAxisRenderer.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/XAxisRenderer.java
@@ -207,7 +207,7 @@ public class XAxisRenderer extends AxisRenderer {
                 if (mXAxis.isAvoidFirstLastClippingEnabled()) {
 
                     // avoid clipping of the last
-                    if (i == mXAxis.mEntryCount - 1 && mXAxis.mEntryCount > 1) {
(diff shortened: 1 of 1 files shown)


longup:
Please review this commit and merge to fix this bug, thank you.

AliGolmirzaei:
I have the same issue and I think this is the fix. please review it. thank you

almic:
This bug is not caused by the XAxisRenderer. The value you changed, `i` to `i / 2` is checking if the current drawn label is the last label, and if so, it correctly shifts it back `width / 2`, since the label is already centered and only half clips off.

Please use `xAxis.setAvoidFirstLastClipping(true)`. If that still doesn't fix the problem, you can simply increase the visible X range with `xAxis.axisMaximum(float)`, just add `0.5f` or similar to the last X value.

longup:
Hello, I have use xAxis.setAvoidFirstLastClipping(true) and increase the visible X range with xAxis.axisMaximum(float), but it crash, like this:
`float lastValue = lineChart.getXAxis().getAxisMaximum();
        lineChart.getXAxis().setAxisMaximum(lastValue + 1.0f);`

And I have try a lot of values, but it doesn't work. Only if I changed  the value  i to i / 2, it worked fine.

Thank you，please review it.

PR #3878, opened 2018-03-21
Prevent spurious garbage collection

marcelelements:
I've noticed that in production the Bitmap occasionally goes missing between being turned into a weakreference and where it is extracted again and passed to the canvas - after all there are no references to the object in the short space between the two lines of code.


MGaetan89:
I guess the solution in #3207 is a bit more elegant, and fixes other issues as well.

almic:
Haven't gotten very far into the issue list, but boy this problem is in a lot of issues!

marcelelements:
#3207 - and some of the others - are basically a larger patch which includes this one verbatim. I leave the question of elegance to the observer, but either that patch or this one should deal with the crashes I've observed.

almic:
Pulled 3702, closing this one.

Issue #3895, opened 2018-03-28
PieChart: Option for setting min angle for pie slices


sembozdemir:
![PieChart](https://i.snag.gy/v3N6iM.jpg)

Is there any way to set minimum draw angle for pie slices? Values are 0.2%, 0.8%, 1%, 2%, 16%, 80% here. But, I want to see slices which are less than 5% the same size. I mean percentage of slices should be like 5%, 5%, 5%, 5%, 16%, 64%(rest of space).

Or, do you suggest any workaround solution for this?

rzhik:
setMinAngleForSlices() method will be very useful, when in will be available in new release?

almic:
This is definitely going to added in the next release, but I cannot give a timeline for it yet. A lot of the features planned are user-made, and have some outstanding requests for changes before they can be merged.

amit0021:
I Want to display on Pie Char only. It is possible with this library.

PR #3896, opened 2018-03-28
PieChart: Add option for setting min angle for pie slices

sembozdemir:
Fixes: #3895

When set `pieChart.setMinAngleForSlices(18f);`, minimum slice will take up %5 at least (for full pie-chart). It is a solution for overlapping issues and small values.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/charts/PieChart.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/charts/PieChart.java
@@ -94,6 +94,8 @@ public class PieChart extends PieRadarChartBase<PieData> {
 
     protected float mMaxAngle = 360f;
 
+    private float mMinAngleForSlices = 0f;
+
     public PieChart(Context context) {
         super(context);
     }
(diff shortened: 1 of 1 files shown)


almic:
I suggest doing this from a different method. Instead of doing this in PieChart.java, do this in the renderer and make it default to `0.5f`. That way the renderer is doing what it's supposed to do, render the PieChart slices.

RobertZagorski:
Change looks very good.
It might be nice, if it would be shown in sample app. You might add it as menu option in sample app. What do you think @sembozdemir ?

bksinha4497:
pieChart.setMinAngleForSlices(18f);
For me setMinAngleForSlices() function is not available  in com.github.PhilJay:MPAndroidChart:v3.0.3'

almic:
@bksinha4497 are you kidding me? This is not merged, so obviously this is not available yet.

rzhik:
I have problem in my project. When slice is too small whole pie chart is colored in that slice color. This method can solve my problem. If it will not be available in near future, can you suggest me other solution?

sembozdemir:
@almic Sorry for late answer. I was on vacation.

I have checked your suggestion. For me, it is better to calculate angles and make them ready before rendering operation. Because, we have to calculate several times (drawDataset, drawHighlighted... methods) if we do this in renderer. Renderer looks like being responsible for rendering pie chart slices with given angle values. So, it is ok that angle values are ready for rendering before they are arrived to renderer.

almic:
@sembozdemir Thank you for your contribution! Although I didn't use your exact code in favor of a more functional approach when min angles are too big, I would like to credit you for bringing this to attention and providing the base for this feature.

Added in aea2ff3417e30d6d4b1ce7e777cbd8bc83e1c95d

Issue #3897, opened 2018-03-28
PieChart: Use slice color as value line color


sembozdemir:
![PieChart](https://i.snag.gy/zrOmTf.jpg)

I want to get this result but only one color can be set for value lines. It would be better to have an option to use slice color as value line color.

sembozdemir:
I added this method. Look at my [PR](https://github.com/PhilJay/MPAndroidChart/pull/3898).

luckerbai:
Is there any method just show one value line when clicked, not show all values lines? thank you.

hxkdidi:
how can i get this code from your pr? @sembozdemir

abdulBasitWaqas07:
Can you please share the full code? by using that code you are showing piechart like this? thanks in advance

PR #3898, opened 2018-03-28
PieChart: Add option for using slice color as value line color

sembozdemir:
Fixes: #3897

If `dataSet.setUsingSliceColorAsValueLineColor(true);` is called, value line color will be the same color with slice.


almic:
Looks great. I would suggest going all the way with your idea on the value lines ending outside the circle with a dot at the end, but this works fine the way it is. If you could add the option to draw lines with an entirely separate color list, that'd be awesome too.

abdulBasitWaqas07:
how can i add the value line on the basis of each slice?

PR #3971, opened 2018-04-23
draw backbround for y value of line chart

xiepeijie:
Fix draw icon not deformation , 
draw backbround for y value of line chart .


almic:
The current implementation works just fine. And if you want to add this new image, you may as well make it compatible for all dpi (mdpi to xxxhdpi). And you should also make it optional, the current implementation forces it on all values, regardless of highlighting.

Basically, make it optional and only for highlighted values. See the MyMarkerView class in the example project. You can technically already do what you are trying to do, just setting a different background in the custom_marker_view.xml.

PR #3975, opened 2018-04-25
update to Android Studio 3.1.2

hannesa2:
This PR updates to most recent stable Android Studio 3.1.2

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Mon Nov 20 11:59:54 CET 2017
+#Wed Apr 25 08:04:33 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
--- a/MPChartExample/build.gradle
+++ b/MPChartExample/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'realm-android'
 
 android {
     compileSdkVersion 27
-    buildToolsVersion '26.0.2'
+    buildToolsVersion '27.0.3'
     defaultConfig {
         minSdkVersion 16
         targetSdkVersion 27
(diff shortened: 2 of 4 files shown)


MPChartExample/src/com/xxmassdeveloper/mpchartexample/notimportant/DemoBase.java
```diff
@@ -36,6 +37,8 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
         mTfRegular = Typeface.createFromAsset(getAssets(), "OpenSans-Regular.ttf");
         mTfLight = Typeface.createFromAsset(getAssets(), "OpenSans-Light.ttf");
+
+        Log.d("OpenActivity", this.getClass().getName());
```
MGaetan89: I guess there is no reason for this change either (and its import).

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Nov 20 11:59:54 CET 2017
+#Wed Apr 25 08:04:33 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
MGaetan89: The latest version of Gradle is `4.7`.

hannesa2: gradle build tools 3.1.2 doesn't work with gradle > 4.5

hannesa2:
Sorry there were to much local changes in the PR. I changed this.
Btw, a lot of other things could be changed, but this is out of scope here
eg:
image

almic:
It does seem that new gradle versions are available, however since Android Studio is only automatically updating to 4.4, that should be fine for now. I just built this version and it worked, merging.

hannesa2:
gradle build tools 3.1.2 doesn't work with gradle > 4.5 

> Failed to apply plugin [id 'com.android.application']
   > Configuration on demand is not supported by the current version of the Android Gradle plugin since you are using Gradle version 4.6 or above. Suggestion: disable configuration on demand by setting org.gradle.configureondemand=false in your gradle.properties file or use a Gradle version less than 4.6.

First suggestion doesn't work too. 
If you have a better way, let us know

PR #3992, opened 2018-05-02
BUG:setGranularity is not work

Aslanchen:
when i use this infomation：

```
XAxis xAxis = mChart.getXAxis();
xAxis.setGranularity(60);
mChart.setVisibleXRange(180, 180);
```
data count ：setData(1440)
and set
`mChart.setVisibleXRange(180, 180);`

the setGranularity is not work，always 100.
I find the AxisRenderer.class line 176 code:

```
       if (intervalSigDigit > 5) {
            // Use one order of magnitude higher, to avoid intervals like 0.9 or
            // 90
            interval = Math.floor(10 * intervalMagnitude);
        }
```
I update the 5 to 6，it is OK。
so，is this the solution？


almic:
I take it you'll see my reply on your issue, so I'm closing this.

Issue #1604, opened 2016-03-20
ScatterDataSet.copy() misses some relevant values


vit001:
For example, method should include

```
copied.setDrawValues( isDrawValuesEnabled() );
```

Without it, the mDrawValues is lost during the copy.

PhilJay:
@danielgindi I have been thinking about dropping the "copy" feature for a long time. What do you think?
Should we keep it or just remove it from the library?

danielgindi:
It seems like people are using it... So I'm not really sure about this :)

PhilJay:
Yea I know :-)
But this bug has been there since the beginning, and happened to be unnoticed until now :-)

almic:
Fixed by above mentioned PR

PR #4008, opened 2018-05-08
Update and reorganise copy data sets methods (Fix for #1604).

RobertZagorski:
Issue has already been created where the problem: [#1604]

## PR Description
Copying data sets methods were missing some fields. I refactored the way data sets are copied and added missing fields.


almic:
Nice work, looks fine from here.

Issue #2369, opened 2016-10-17
LineChart: Axis mLabelRotatedHeight is counted twice when calculating offsets


matiash:
It looks like the legend adds this height to its own offset. This happens in every case, but it's most noticeable when the labels are long, since the space taken is bigger. You can clearly see it (blank space between chart and legend) when comparing these two screenshots

<img src="https://cloud.githubusercontent.com/assets/357354/19453582/4607586a-948c-11e6-9466-05a878bc22f6.png" width="300"> and <img src="https://cloud.githubusercontent.com/assets/357354/19453584/4611ca66-948c-11e6-8711-4e36664abcfe.png" width="300">

(This was tested with the latest commit, as of today)

**Steps to reproduce**

Edit `LineChartActivity1.java` in the sample project, and add the following lines to the onCreate() method:

    xAxis.setLabelRotationAngle(-90f);
    xAxis.setValueFormatter(new IAxisValueFormatter() {
        @Override
        public String getFormattedValue(float value, AxisBase axis) {
            return "A long text for value" + (int)value;
        }

        @Override
        public int getDecimalDigits() {
            return -1;
        }
    });

I believe the bug is in `BarLineChartBase.calculateLegendOffsets()` which adds `getXAxis().mLabelRotatedHeight` to the top offset of the legend (or bottom, depending on its position):

    if (getXAxis().isEnabled() && getXAxis().isDrawLabelsEnabled())
        offsets.bottom += getXAxis().mLabelRotatedHeight;

I tried commenting out those lines and the problem goes away, but I'm not sure if there could be any other repercussions.

PhilJay:
Thank you for this very detailed issue description and reproduction guide. If only every issue would be as well formulated as this one.

RobertZagorski:
Issue can be closed, PR https://github.com/PhilJay/MPAndroidChart/pull/4013 for that was merged.

PR #4013, opened 2018-05-09
Remove mLabelRotatedHeight counted twice, when calculating legend offset

RobertZagorski:
Fix for #2369.

## PR Checklist:
Modified the code as described in Issue #2369 to see, that the issue still exists.
Checked against different positions of Legend (TOP, BOTTOM).
No need for that, I think.
- [x] If this is a significant change, an issue has already been created where the problem / solution was suggested
Issue #2369

## PR Description
Removed statements described in ticket as they were completely not needed. Calculating offsets is already done in `BarLineCharBase#calculateOffsets(...)` (`calculateLegendOffsets(...)` is called inside it).
The code, where calculating legend offset is done in [`BarChartLineBase#473`](https://github.com/PhilJay/MPAndroidChart/compare/master...RobertZagorski:bugfix/%232369-legend-offset-is-not-calculated-twice-for-line-chart?expand=1#diff-c2688ad7111a70de6e0354d19bbff337R473)
Modified temporarily `LineChartActivity1` code to reproduce the issue and tested against various legend positions.
Additionally I changed one inner variable name to be more of Java style.


almic:
LGTM

And thank you for using the template 👍

PR #4018, opened 2018-05-10
[Enhancement] Remove redundant findViewById casts in example application

RobertZagorski:
## PR Checklist:
Yep, opened the example app, clicked through it to see, if views initialise properly.
- [X] I have added/updated examples and tests for any new behavior.
Modified existing ones.
- [X] If this is a significant change, an issue has already been created where the problem / solution was discussed
This was suggested in [PR 3975](https://github.com/PhilJay/MPAndroidChart/pull/3975) that migrated the project to newest Gradle version and Gradle version plugin.

## PR Description
After migration to API 26, new version of `findViewById` method was added that caused, that casts to specific view instance is unnecessary. This PR removes those unnecessary casts.


almic:
Awesome, must've been a little tedious!

PR #4086, opened 2018-06-06
添加了一个设置圆角柱状图的方法

wardenlzr:
## PR Checklist:


wardenlzr:
添加了一个设置圆角柱状图的方法

Issue #4024, opened 2018-05-10
Entry(x,y) doesn't handle correctly x as milliseconds (unix time)


pwireless:
Entry(x,y) doesn't handle correctly x as milliseconds (unix time)

Did the following simple lines: 

            long l1 = System.currentTimeMillis();
            long l2 = l1 + 500;
            Entry e1 = new Entry(l1, 10);
            Entry e2 = new Entry(l2, 10);
            long e1long = (long)e1.getX();
            long e2long = (long)e2.getX();
            Log.d("MEEE2", String.format(Locale.US, "l1 = %d -- e1long = %d,    l2 = %d -- e2long = %d", l1, e1long, l2, e2long));

and the result was :

l1 = 1525979109063 -- e1long = 1525979152384,    
l2 = 1525979109563 -- e2long = 1525979152384

i.e. the Entry doesn't handle milliseconds correctly, due to the float being 32-bit FP format.

The LineChartTime class in the MPAndroidChart example, has the example with hours, which then converts to milliseconds in the value formatter. It would be great to have the example with milliseconds instead of hours, as it is more generic.

Entry should be able to accept correctly milliseconds.

A new Entry constructor would be ideal : 

Entry e  = new Entry(long x, float y);

(in the previous library versions, the addXValue would save the situation)

almic:
This is a well known problem, and the only solution is to just not use such big numbers. There are better ways to handle times! Use smaller integer X and Y values, and apply the `AxisValueFormatter` and `ValueFormatter` classes. You can pass the original data set and map the times 1:1

Also, the suggestion to [convert to double values](https://github.com/PhilJay/MPAndroidChart/issues/2891) is a contender to fixing this, although this method is still not the best solution. I would almost insist on allowing you to pass a `Calendar` or `Date` object—and let the DataSet class reformat them into intervals with simplified indexes—before letting developers get the renderer tangled up with huge double values.

As I always suggest to anyone having issues with big numbers in this library, cut them down into smaller and more meaningful values! Converting the base number `1802210870` into `1802` and then value format that into `1.8B` or `2B` is much better for the end-user! Same with dates and the developer, pass small indexes spaced according to your time interval, and use the formatter to turn them into meaningful dates and times by passing the original timestamps into it.

Issue #3770, opened 2018-01-30
Data reduction issue - data has been down sampling when render with LineChart


yangdong:
I am using LineChart to display **time based data** collected from somewhere **every 5 seconds** on a daily basis, which means my chart will be maximum about **17000+ points.**

However, I found LineChart won't render the whole points, instead, it will **down sample my data to about 2 minutes**, and the entry still **don't render** after I **zoom in** the chart.

I've go through the MPAndroidChart docs and your examples, but no luck, and I think this may related to #7 .

![image](https://user-images.githubusercontent.com/954058/35562188-9548ebb8-05ed-11e8-9f11-fce3472cd7b4.png)

JStav:
Perhaps using axis.setGranularityEnabled(true) and axis.setGranularity will help. Also chart.setVisibleXRange to something smaller than the full dataset. Showing less data and allowing it to scroll could also help.

yangdong:
@JStav thanks for your reply. I think I might found the issue.

**Having Issue If:**
My data is time series data, value of x is unix timestamp new ```Date().getTime()```, and then I use date valueFormatter to format the xAxis to date.

**No Issue if**
Update the x value to a smaller number ```Date().getTime() - ${unixtimeOfStartOfThatDate}```, all data will be displayed.

yangdong:
this is not related to this repo, will close it

timusus:
If you're using a unix timestamp, the problem is when it gets converted to Float, it loses precision, so lots of timestamps that actually represent different times end up being rounded and represented as the same x value. See #2891

Issue #3630, opened 2017-12-05
Not able to draw a proper Linear graph where X is date timestamp


eurbon:
This is how my graph looks when I use graphView lib: http://prntscr.com/hjg501

Hellochart: http://prntscr.com/hjg8fa

MPChart: http://prntscr.com/hjghbb

My dataset on both graphs is this (X, Y):

1512488280000	1.200000048
1512488310000	1.200000048
1512488346000	1.200000048
1512488370000	3.599999905
1512488400000	1.200000048
1512488430000	1.200000048
1512488460000	1.200000048
1512488490000	1.200000048
1512488524000	1.200000048
1512488550000	6
1512488580000	1.200000048
1512488612000	1.200000048
1512488646000	1.200000048
1512488674000	3.599999905
1512488702000	1.200000048
1512488730000	1.200000048
1512488760000	6
1512488790000	1.200000048
1512488820000	1.200000048
1512488850000	1.200000048
1512488880000	2.400000095
1512488910000	1.200000048
1512488940000	1.200000048
1512488970000	1.200000048
1512489000000	1.200000048
1512489030000	1.200000048
1512489060000	12
1512489090000	1.200000048
1512489126000	13.19999981
1512489150000	7.199999809

I see that in hellochart and mpchart graphs look the same, while in graphview it shows properly, so I must be doing something wrong then...

Is there a way to make it work?

timusus:
If you're using a unix timestamp, the problem is when it gets converted to Float, it loses precision, so lots of timestamps that actually represent different times end up being rounded and represented as the same x value. See #2891

khushbushah2302:
Hello, any update on this ? 
I have the same issue and I am looking for a solution to plot a lot of points where X is unix timestamp.

Issue #3323, opened 2017-08-08
Time series plotting?


karanrajpal14:
How would I be dealing with plotting time-series data?

I'm just trying to plot some dummy data using:
```
double x = 1501572740
int y = 25 
```
And then plot it using a thread that calls addEntery2() in onResume() like so:
```
protected void onResume() {
		super.onResume();
		new Thread(new Runnable() {
			@Override
			public void run() {
				runOnUiThread(new Runnable() {
					@Override
					public void run() {
						for (int i = 0; i < 10; i++) {
							addEntry2();
						}
					}
				});
				try {
					Thread.sleep(600);
				} catch (InterruptedException e) {
					e.printStackTrace();
				}
			}
		}).start();
	}
```
which then calls addEntry2():
```
private void addEntry2() {

		LineData data = chart.getData();

		if (data != null) {

			ILineDataSet set = data.getDataSetByIndex(0);

			if (set == null) {
				set = createSet();
				data.addDataSet(set);
			}

			Log.d("TAG", "addEntry2: x" + this.x);
			Log.d("TAG", "addEntry2: y" + this.y);

			data.addEntry(new Entry((float) this.x, (float) this.y), 0);
			++this.x;
			++this.y;
			data.notifyDataChanged();

			Log.d("TAG", "addEntry2: new x" + this.x);
			Log.d("TAG", "addEntry2: new y" + this.y);
			
			chart.notifyDataSetChanged();
			chart.setVisibleXRangeMaximum(120);
			chart.moveViewToX(data.getEntryCount());

		}
	}
```
Now, the problem here is that it just plots one point and then nothing happens. But, if I reduce the number of digits in the timestamp, it works just fine but that won't help me. Is there something that I'm missing out on? Any way I can get around this? I'd appreciate some help here as I've been racking my brain for a long while with no results whatsoever.
Thanks in advance!

timusus:
If you're using a unix timestamp, the problem is when it gets converted to Float, it loses precision, so lots of timestamps that actually represent different times end up being rounded and represented as the same x value. See #2891

PR #4121, opened 2018-06-27
Accuracy improvement

devnokiyo:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A]

## PR Description

Nice to meet you, my name is devnokiyo. I will Pull Requests for the first time.

If the value set for DataEntry is Float type, the possibility of rounding error of the value displayed in the chart increases.
I modified it to Double type like iOS's Charts.

change.

Yes.

### DIFF
--- a/MPChartExample/src/com/xxmassdeveloper/mpchartexample/BarChartActivityMultiDataset.java
+++ b/MPChartExample/src/com/xxmassdeveloper/mpchartexample/BarChartActivityMultiDataset.java
@@ -93,7 +93,7 @@ public class BarChartActivityMultiDataset extends DemoBase implements OnSeekBarC
         xAxis.setCenterAxisLabels(true);
         xAxis.setValueFormatter(new IAxisValueFormatter() {
             @Override
-            public String getFormattedValue(float value, AxisBase axis) {
+            public String getFormattedValue(double value, AxisBase axis) {
                 return String.valueOf((int) value);
             }
         });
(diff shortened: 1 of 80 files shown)


timusus:
Resolves #4024

timusus:
Resolves #4121, #4138, #3630, #3323, #3138, #3088, #2348, #3770, #2327 & #2195

timusus:
@devnokiyo thanks for attempting to solve this issue.

Unfortunately, your PR doesn't really solve the issue of this library's reliance on `floats` - you have changed `float` to `double` in a bunch of places, but then in various places these `doubles` are being cast back to `float`. This is a 'lossy conversion', meaning that the original `double` data is lost, as it cannot be accurately stored in a `float`.

A proper solution to this problem will require completely eliminating doubles. Unfortunately, there are some libraries (such as `android.graphics.matrix`) which don't support `double`, so these would have to be replaced.

As is, this change just masks the problem to some degree. The charts will still render data inaccurately if that data is represented by more than 6-7 digits.

devnokiyo:
Thank you for checking the source changes.
As for android.graphics.matrix concerning drawing certainly it is excluded from the subject of correction.
This PR only increases the accuracy of the numerical value that displays the value of the graph.
When I displayed the monetary amount with my application, I got an error and PR.
I will wait for a fundamental solution.

timusus:
I'm not a maintainer of this library, but I would argue that this PR probably can't be accepted, as it doesn't fix the problem correctly. It actually introduces errors, as it allows users to input `double` values and then silently performs a lossy conversion which causes those `doubles` to become inaccurate.

devnokiyo:
Since this matter will not proceed, I am closing it.

devnokiyo:
Since this matter will not proceed, I am closing it.

hannesa2:
Just FYI: I tried to make it generic (https://github.com/AppDevNext/AndroidChart/pull/684), but unfortunately it's gone dormant....

PR #4162, opened 2018-07-25
Providing Real Slices instead of a whole TransparentCircle

cadicetienne:
## PR Checklist:


cadicetienne:
Hello ! Any news from the PR ?

In our integration of this nice library we had to provide proper slices with each its own inner arc, using the PieChart.

The current implementation is to draw a whole circle (mTransparentCircleRadiusPercent), which is inappropriate if you provide space between slices (then you see the circle go over the background). It isn't easy to find out because all the examples provided have a white background & the semi transparent circle has a white base color, so you might think the 'transparentCircle' runs only above the slices. But if are up to try and use it with another background color then it becomes obvious that something is wrong !

We would love to see the PR accepted and be able to still use the official library instead of our own fork.

Thank you for your time !

PR #4169, opened 2018-07-31
Android studio3.1.3

hannesa2:
* simply update to most recent stable build tools version
* keep Gradle version and add result of run
`./gradlew wrapper`


almic:
Again, thanks! I'm sure most of this was added in d67ea481af1e8528c617dfd404d0c7827b0134b2 but again tell me if I missed something.

PR #4170, opened 2018-07-31
fix lint issues

hannesa2:
fixes some Lint issues and add partial codeformat


almic:
Thank you, looks like I just did most of this in d67ea481af1e8528c617dfd404d0c7827b0134b2. Let me know if there was something I missed though.

PR #4172, opened 2018-07-31
everyone use same codeformat now

hannesa2:
when someone touch a file during a PR a lot of changes are done automatically, because everyone uses individual codeformat and a **search for significant changes is a nightmare**.
To prevent this, I made this PR.
When someone clone this repo and make a PR he uses now same codeformat like every other developer.
What's missing : someone (a maintainer) has to perform this  
image

I could do this as well, but then I change every file and (I guess) this PR will be never merged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 bin/
 gen/
 generated/
+docs/
 finalOutput/
 projectFilesBackup/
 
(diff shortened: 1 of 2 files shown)


almic:
Thanks for this, I'll get back in a few days and probably pull this. If you don't hear anything from me in the next few days, please ping me by mentioning me :)

hannesa2:
"please ping me by mentioning me" 
@almic here is the ping

PR #4174, opened 2018-08-01
test

churikaustubh:


almic:
I'm sorry, but from the name of the commits it looks like you've recreated the Candlestick chart, which already exists in this library. If you have any features or functionality you want to add to the existing code, then please do so, but I cannot accept a copy of the existing chart.

PR #4198, opened 2018-08-16
Fix LargeValueFormatter

mtrakal:
On Android 8+ is format: 1e1, 1e2 1e22 for Slovak language although that in constructor is set: `###E00`

When you will have more than 2 numbers after `E` original regex will not work, because take just first 2 numbers.

## PR Checklist:

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/LargeValueFormatter.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/LargeValueFormatter.java
@@ -88,7 +88,7 @@ public class LargeValueFormatter implements IValueFormatter, IAxisValueFormatter
         int numericValue2 = Character.getNumericValue(r.charAt(r.length() - 2));
         int combined = Integer.valueOf(numericValue2 + "" + numericValue1);
 
-        r = r.replaceAll("E[0-9][0-9]", mSuffix[combined / 3]);
+        r = r.replaceAll("[Ee][0-9]*", SUFFIX[combined / 3]);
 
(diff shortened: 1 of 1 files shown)


mtrakal:
https://regex101.com/r/Voqm04/1

mtrakal:
using:
```
   if (combined / 3 < SUFFIX.length) {
            r = r.replaceAll("[Ee][0-9]*", SUFFIX[combined / 3]);

            while (r.length() > MAX_LENGTH || r.matches("[0-9]+\\.[a-z]")) {
                r = r.substring(0, r.length() - 2) + r.substring(r.length() - 1);
            }
        }
```

When you have number bigger than exponents in `SUFFIX`  it crashed on ArrayIndexOutOfRange, so it should be in if statement...

cpboyd:
Would it be better to use `[Ee][0-9]+` since there should probably be at least 1 number following the exponent?

mtrakal:
@cpboyd yes, it could be better use `+` instead of `*`

Deepakkast:
not getting what i want to achive

Deepakkast:
public float getPerStu(int total, int per) {
    Logger.debug( "get Per stu into Total is " + total + "And per is:" + per );
    float percentage = 00;
    if (total != 0) {
        Logger.debug( "into if: " + total );
        Logger.debug( "per is: " + per );
        percentage = (100 * per) / total;
        Logger.debug( "percentage is: " + percentage );
        return percentage;

almic:
@Deepakkast this is not the place to get programming help, please do not post code without any real explanation of why. Open an issue and _follow the templates_ if you have a problem with the library

Issue #4214, opened 2018-09-02
Changing xAxis label count with data set change results in old dataset values passed to formatter


deividasstr:
**Summary**
Changing xAxis label count with data set change results in old data set values passed to formatter, but not always. 
Setting of params on the first data set:
``` 
with(xAxis) {
        position = XAxis.XAxisPosition.BOTTOM
        this.typeface = typeface
        setDrawGridLines(false)
        this.labelCount = labelCount
        valueFormatter = xAxisFormatter
    }
...
barChart.data = barData
```
on the first load everything is ok.
Then I refresh the data, formatter and label:
```
...
    val xAxisFormatter = data.xAxisFormatterByRange()
    val labelCount = data.xAxisLabelCountByRange()

    barChart.xAxis.labelCount = labelCount
    barChart.xAxis.valueFormatter = xAxisFormatter

    val dataSet = view.data.getDataSetByIndex(0) as BarDataSet
    dataSet.values = entries

    barChart.data.notifyDataChanged()
    barChart.notifyDataSetChanged()
```
The first time the data refresh happens, everything goes well.
I get the crash on the second refresh on line `barChart.notifyDataSetChanged()` because values from old data are passed to a new formatter.

When I remove the setting of labelCount from both the init and refresh, everything works well.
When I leave the labelCount setting only in the init or refresh, I get the same crash.

I need this as I use date ranges. For the period of week, all 7 labels are present without any labelCount set, but for periods of month and year, default 6 labels are not good enough. Label counts are 7, 15 and 12.

**Expected Behavior**
   After `barChart.notifyDataSetChanged()`  chart is refreshed with new data, formatter and label count.

**Possible Solution**
mEntries in XAxis are not always refreshed on labelCount changes? I could not find where they are set or reset.

**Device (please complete the following information):**
 - Device: Google Pixel 2
 - Android Version 9
 - Library Version  3.0.3

Other bar chart settings: 
```
        barChart.setDrawValueAboveBar(true)
        barChart.setScaleEnabled(false)
        barChart.legend.isEnabled = false
        barChart.description.isEnabled = false
        barChart.setNoDataText(noDataText)
        barChart.setNoDataTextTypeface(typeface)
```

Crash log:
```
org.threeten.bp.DateTimeException: Invalid value for DayOfWeek: 8
        at org.threeten.bp.DayOfWeek.of(DayOfWeek.java:144)
[...]

PR #4217, opened 2018-09-04
Fix issue #4214

deividasstr:
setting xAxis label count with data set changes results in old dataset values passed to formatter

## PR Checklist:
- [ x] I have tested this extensively and it does not break any existing behavior.
       

## PR Description

Not always recreating entries array caused this bug.

It only removes `if (mAxis.mEntries.length < n) {` check, always recreates the array 

Because it fixes the unwanted behavior.


damiafuentes:
Why's no-one merging this?

hannesa2:
Because it's a dead repo. I at least helped myself and merged a lot of open pull requests 
https://github.com/AppDevNext/MPAndroidChart/pull/70

PR #4241, opened 2018-09-25
Move to Jetpack libraries (androidx)

defer:
## PR Description
Jetpack (androidx 1.0.0) was finally released as stable and libraries should start moving to it.

No behavior changes. In fact, the use of support library is restricted to annotations in the chart library, most of the changes are actually to the example app.

It ensures that we can continue using Jetpack as it gets developed instead of using the soon-to-be stagnant support library.


almic:
Thank you for doing this! I just implemented this however in d67ea481af1e8528c617dfd404d0c7827b0134b2. I'm sorry that you won't be properly credited for this, but that you nonetheless!

PR #4309, opened 2018-11-14
Highlight by index

asherali:
## PR Checklist:


almic:
@asherali could you please test this to make sure it works? Just by looking at the code you wrote a function but it isn't actually called anywhere, this should at least be used in one of the examples.

PR #4310, opened 2018-11-15
Remove redundant code

luuillu:
## PR Checklist:
       

## PR Description

remove a piece of useless code

remove some line from CharData.java

the code will be more concise.


almic:
@luuillu thanks for doing this, but it may be a few more days until I pull it. Hold tight!

Issue #4360, opened 2018-12-23
BarLineChart assumes data is ordered


varunrau:
**Summary**
BarLineChart assumes values are ordered. `DataSet.getEntryIndex` does a binary search over values to find indices, but values isn't necessarily sorted. This can also possibly cause NPEs later in the program. In my use case, the order of the values needs to be preserved, so sorting them before rendering is incorrect.

**Expected Behavior**
No NPE on unsorted input data. 

```
12-18 22:01:24.353 30855 30855 E AndroidRuntime: java.lang.NegativeArraySizeException: -16
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at com.github.mikephil.charting.utils.Transformer.generateTransformedValuesLine(Transformer.java:178)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at com.github.mikephil.charting.renderer.LineChartRenderer.drawValues(LineChartRenderer.java:547)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at com.github.mikephil.charting.renderer.CombinedChartRenderer.drawValues(CombinedChartRenderer.java:96)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at com.github.mikephil.charting.charts.BarLineChartBase.onDraw(BarLineChartBase.java:264)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.View.draw(View.java:20207)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.View.updateDisplayListIfDirty(View.java:19082)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4317)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4290)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.View.updateDisplayListIfDirty(View.java:19042)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4317)
12-18 22:01:24.353 30855 30855 E AndroidRuntime: 	at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4290)
[...]

PR #4363, opened 2018-12-24
Remove assumption that input data is ordered

varunrau:
The original implementation of `getEntryIndex` was using a binary
search. This relies on the input data being sorted. This commit replaces
that implementation with a linear search, which does not have that
requirement. This allows users to plot unordered data.

## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A, but I did create an issue here: #4360

## PR Description
The original implementation of `getEntryIndex` was using a binary
search. This relies on the input data being sorted. This commit replaces
that implementation with a linear search, which does not have that
requirement. This allows users to plot unordered data.


varunrau:
I tested this with the example app and didn't see a significant slowdown when using 500 points in the linear chart.

PR #4378, opened 2019-01-05
introduce a consumer proguard file

hannesa2:
If there is a need for proguard, this is the way to go.
Recent proguard settings are useless


MGaetan89:
This is a duplicate of #2596

PR #4450, opened 2019-03-09
Fix Bubble Highlight

regas99:
## PR Checklist:
- [ X] I have added/updated examples and tests for any new behavior.
- [ ] If this is a significant change, an issue has already been created where the problem / solution was discussed:

## PR Description
Highlighting a Bubble Chart has a number of problems. Sometimes the correct
bubble is highlighted, sometimes no bubble is highlighted, and sometimes
the wrong bubble is highlighted.

The root cause of this behavior has three components:

1) Despite the variety of chart types supported, the underlying architecture
assumes that, for the x axis, entries are ordered and distinct (non-overlapping).
A Bubble Chart violates both of these paradigms - bubbles can have the same x but different
y's, and closely spaced (x-axis) entries can overlap. The application attempts
to address the former (see #3 below), but has no mechanism to handle the latter.

2) The mSize field in Bubble Entry has no concrete meaning. It is correlated
to the size of the rendered circle, but it first undergoes a transformation.
The transform scale factor changes every time the map is zoomed,
and may change when a new entry is added (ala realm.io). Regardless, it
is re-computed every time the chart is rendered.

3) DataSet#getEntryIndex has a bug. It properly detects when there are
multiple entries with the same x value (equal, but does not handle
overlapping) and puts those in a list. When searching that list for
the closest y value, only the first or second list element is
returned - never the third, fourth, etc.

The changes are pretty well documented in the commit notes.

There is more clean-up that could be done. For example,
ChartHighlighter still exhibits issue #1 above, and only considers
bubbles immediately above or below the touch point. Click underneath
a vertical column in the demo to see this behavior.

I believe this pull request should close issues 3102 and 2707. It
also duplicates the fix in 3463.


eggham0518:
I took your all files and changed it all, but it still doesn't click exactly.
 Especially when the size is very different.    

Is it impossible for this library to be clicked exactly like the bubble chart on this link?
https://github.com/AnyChart/AnyChart-Android

regas99:
Yes you can. I created a quick mockup in a separate repository [here](https://github.com/regas99/MPAndroidChart_SportsmanDemo) .

sportsman

eggham0518:
Thank you for your answers.
But I think there was a communication problem.

I just want the exact click range like AnyChart.
It is still selected by clicking on the space or by touching near the circle.

It's a fairly famous library and I can't understand why these bugs still exist. Is this a problem that no one can fix? I'm sorry to ask you a sensitive question.

regas99:
You are correct - your first question was very poorly stated. So is your second. Exactly what behavior do you see versus what behavior do you want to see?

eggham0518:
I am not used to writing English, so I think I misrepresented it. I'm sorry.

AnyChart responds to touch with the range of bubbles drawn, 
while the mpAndroidChart responds with the nearest bubble from the touch.

This method that the mpandroidchart is using cannot touch bubbles accurately
when they are large or overlap each other.

PR #4567, opened 2019-06-07
Clarify Javadoc for getOffsetForDrawingAtPoint()

avianey:
posY indicates Y position

## PR Checklist:
- [N/A] I have tested this extensively and it does not break any existing behavior.
- [N/A] I have added/updated examples and tests for any new behavior.
- [N/A] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A, or add link to issue here]


SubhrajyotiSen:
Merged this at https://github.com/SubhrajyotiSen/MPAndroidChart which will be an updated fork

PR #4624, opened 2019-07-27
Removed G+ Button

locked-fg:
G+ is no more. Just removed the G+ Share-Button / Link

## PR Checklist:

## PR Description
Removed the G+ Button as GogglePlus is shut down

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,12 +125,13 @@ Please read the [**documentation**](https://weeklycoding.com/mpandroidchart/) fi
 
 <br/>
 
-<h2 id="social">Social Media :fire:</h2>
+<h2 id="
+
+">Social Media :fire:</h2>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,12 +125,13 @@ Please read the [**documentation**](https://weeklycoding.com/mpandroidchart/) fi
 
 <br/>
 
-<h2 id="social">Social Media :fire:</h2>
+<h2 id="
```
MGaetan89: This looks wrong

MGaetan89:
I think you should also remove the `googleplus_icon.png` file

locked-fg:
Damn Right! Removed the icon and put the css-id back in (no clue how/why I removed it)

PR #4625, opened 2019-07-27
Add multiline support at Renderer.drawValue

koji-1009:
## PR Checklist:
  - There is no impact if the user is not using "\n"
  - Following screenshot. If need to add new ValueFormatter and option menu, I will add it.
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: #1507 #1508 #1570

## PR Description

Update drawMultilineText method to show multiline labels.

- Fix `drawMultilineText` , add `isMultilineText` method.
- Change `drawValue` method following classes
  - BarChartRenderer
  - BubbleChartRenderer
  - CandleStickChartRenderer
  - HorizontalBarChartRenderer
  - LineChartRenderer
  - RadarChartRenderer
  - ScatterChartRenderer

- Now, we can't use multiline text. So, in the case of a long label, it will overlap the bar next to it.

ScreenShot

0 | 90
:---: | :---:
 |

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/BarChartRenderer.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/BarChartRenderer.java
@@ -6,6 +6,7 @@ import android.graphics.Paint;
 import android.graphics.RectF;
 import android.graphics.drawable.Drawable;
 import android.text.TextPaint;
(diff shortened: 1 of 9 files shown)


02gaurav:
This seems to be not working. I have tried the same if the XAxis label having 2 lines then it overlaps the chart.

PR #4671, opened 2019-09-28
Update build.gradle

nguyenducviet1999:
## PR Checklist:


MPChartExample/build.gradle
```diff
@@ -22,5 +22,5 @@ android {
 dependencies {
     implementation "androidx.appcompat:appcompat:1.0.2"
     implementation 'com.google.android.material:material:1.0.0'
-    implementation project(':MPChartLib')
+    implementation  project(':MPChartLib')
```
hannesa2, line 25: This PR makes no sense at all, it only violates a nice code style

PR #4683, opened 2019-10-14
Support border color and border width for pie chart slices

ndr3:
## PR Checklist:

Screenshot_20191014-140125

- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A
       

## PR Description
Support border color and border width for pie chart slices

Add support for displaying borders around the pie chart slices (custom border width and color, similar to Bar charts).

It looks like there are other people who need this functionality (see for example the discussion from https://stackoverflow.com/questions/38758885/border-around-pie-chart-in-mpchart)


aflorant:
@PhilJay +1 Really need this or my graphic designer will kill me

nissimr:
@PhilJay +1 would help a lot for empty slices with just a stroke around it and transparent fill

PR #4774, opened 2020-01-03
Fix borders being draw behind fill

EduardoWeber:
## PR Checklist:

## PR Description
When you have a chart with a fillColor specified and fillAlpha = 255 the area below the line would be drawn over the border like this:
graphNotFixed

But the correct behavior would be:
graphFixed


EduardoWeber:
Sorry about these new commits, I forgot to change the branch. Reverted all of them

PR #4787, opened 2020-01-23
Feature/catching up to ios

danielgindi:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed

## PR Description

Catches up with features and mainly bugfixes committed on Charts.


danielgindi:
Working on conflicts, I was based on an older `master`.

PhilJay:
Thanks Daniel for taking the time! Anything I can help with?

danielgindi:
@PhilJay No prob :-) I just did some sync both sides to get the "cross platform" part of it to still be reliable...

PR #4839, opened 2020-03-09
If granularity and the number of labels (with force = true) are set, …

NINDZICU:
…granularity setting is ignored.

## PR Checklist:

## PR Description
If granularity is set and the number of labels with force = true is set, granularity setting is ignored.
Now that this and this are set, the values on the axis do not become less granulaity


hannesa2:
I guess that it's already included here https://github.com/AppDevNext/AndroidChart/blob/e09229d6884320f1866a7de9c01db1c9ac5fca67/chartLib/src/main/kotlin/info/appdev/charting/renderer/AxisRenderer.kt#L148

Issue #4225, opened 2018-09-11
Utils drawImage issue


enjoy008:
```java
public static void drawImage(Canvas canvas,
                                 Drawable drawable,
                                 int x, int y,
                                 int width, int height) {

        MPPointF drawOffset = MPPointF.getInstance();
        drawOffset.x = x - (width / 2);
        drawOffset.y = y - (height / 2);

        drawable.copyBounds(mDrawableBoundsCache);
        drawable.setBounds(
                mDrawableBoundsCache.left,
                mDrawableBoundsCache.top,
                mDrawableBoundsCache.left + width,
                mDrawableBoundsCache.top + width);

        int saveId = canvas.save();
        // translate to the correct position and draw
        canvas.translate(drawOffset.x, drawOffset.y);
        drawable.draw(canvas);
        canvas.restoreToCount(saveId);
    }
```

**should change to (width -> height)**
```java
public static void drawImage(Canvas canvas,
                                 Drawable drawable,
                                 int x, int y,
                                 int width, int height) {

        MPPointF drawOffset = MPPointF.getInstance();
        drawOffset.x = x - (width / 2);
        drawOffset.y = y - (height / 2);

        drawable.copyBounds(mDrawableBoundsCache);
        drawable.setBounds(
                mDrawableBoundsCache.left,
                mDrawableBoundsCache.top,
                mDrawableBoundsCache.left + width,
                mDrawableBoundsCache.top + height);

        int saveId = canvas.save();
        // translate to the correct position and draw
        canvas.translate(drawOffset.x, drawOffset.y);
        drawable.draw(canvas);
        canvas.restoreToCount(saveId);
    }
```

almic:
Thanks for pointing this out, adding to the list.

PR #4861, opened 2020-03-28
Fixed width/height bug for drawing images

timbeissinger:
Bug was initially reported Sept 11, 2018: "Utils drawImage issue #4225". PhilJay wrote that he planned to fix it but it hasn't been incorporated yet. (https://github.com/PhilJay/MPAndroidChart/issues/4225)

## PR Checklist:


hannesa2:
Thank you for the bug fix. I apply it here https://github.com/AppDevNext/AndroidChart/pull/638

PR #4866, opened 2020-04-04
Let build and test any pull request with Github actions

hannesa2:
## PR Checklist:
       

## PR Description
Let build and test any pull request


hannesa2:
I see you miss probably this setting
If you enable it, then it looks like this https://github.com/hannesa2/MPAndroidChart/pull/2 

image

PR #4893, opened 2020-05-06
Update BarChartRenderer.java

ShalevCohen2112:
## PR Checklist:


ShalevCohen2112:
l

Issue #3388, opened 2017-08-30
Unable to drag highlight when zoomed into line chart


lesiva:
There doesn't seem to be a way to lock the visible x range to a subset of the graph and then enable dragging of the highlight. 

(60 values in the graph, locked into viewing only the last 12 values -- unable to drag highlight)

paul-freez:
Same thing here. The main problem is that when you force chart to show less values than it has is DataSet - it thinks that is's always zoomed in:
```
boolean shouldPan = !mChart.isFullyZoomedOut() ||
                                !mChart.hasNoDragOffset();
```
Here mChart.isFullyZoomedOut() is always false and we cant highlight values no more

idealllee:
The author's suggestion in https://github.com/PhilJay/MPAndroidChart/issues/948 and is still working for me as long as:
`dragEnabled` is set to `false` and
`highlightPerDragEnabled` is set to `true`. 

@paul-freez from [code](https://github.com/PhilJay/MPAndroidChart/blob/72031d33cacf9d880261e491a5de6f0c1b2de78d/MPChartLib/src/main/java/com/github/mikephil/charting/listener/BarLineChartTouchListener.java#L188) it should be that `shouldPan` logic is bypassed once `dragEnabled` is `false`.

BulatMukhutdinov:
Any progress on this? I need to block scrolling and move highlight while zoomed in

PR #4897, opened 2020-05-08
fix unable to drag highlight when zoomed into line chart

Lloyd0577:
## PR Checklist:
-  I have tested this extensively and it does not break any existing behavior.
-  I have added/updated examples and tests for any new behavior.
-  If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A, or add link to issue here]
       

## PR Description
1.This PR fix the issue that unable to drag highlight when zoomed into line chart related to #3388 .
2.When chart is zoomed, you can drag highlight after chart's `dragEnabled` is set to false.
(In my code, I set `dragEnabled` to false in `onChartLongPressed` callback,then set it to true back in `onChartGestureEnd` callback ).


Lloyd0577:
fix unable to drag highlight when zoomed into line chart

Issue #2071, opened 2016-07-27
First Dataset Using Second's Alpha Value in Line Chart


jngr:
I noticed some odd behavior when using dual line charts. I wanted to have one data series shown as a line and another data series of discrete values with only the data points plotted. Idea was to make the second graph line invisible setting its alpha value to 0. 
- start off e.g. with the dual line chart sample
- use ARGB values for the colors of both graphs
- first graph's mode set to LINEAR
- make the second graph disappear by setting its alpha value to 0
- change first graph's mode to HORIZONTAL_BEZIER or CUBIC_BEZIER
- the first graph and its fill disappear

Playing with different alpha values, it turns out that the first graph's color and fill both use the second graph's alpha value if HORIZONTAL_BEZIER or CUBIC_BEZIER mode are selected. ´
(This is using MPAndrroidChart version 3.0.0 beta.)

**Modified Code from the LineChartActivity2 sample:**

```
            set1 = new LineDataSet(yVals1, "DataSet 1");
            set1.setAxisDependency(AxisDependency.LEFT);
            //set1.setColor(ColorTemplate.getHoloBlue());
            set1.setColor(0xffff0000);
            set1.setCircleColor(Color.WHITE);
            set1.setLineWidth(2f);
            set1.setCircleRadius(3f);
            set1.setFillAlpha(65);
            set1.setDrawFilled(true);
            set1.setMode(LineDataSet.Mode.HORIZONTAL_BEZIER);
            set1.setFillColor(ColorTemplate.getHoloBlue());
            set1.setHighLightColor(Color.rgb(244, 117, 117));
            set1.setDrawCircleHole(false);

            // create a dataset and give it a type
            set2 = new LineDataSet(yVals2, "DataSet 2");
            set2.setAxisDependency(AxisDependency.RIGHT);
            //set2.setColor(Color.RED);
            set2.setColor(0x0000ff00);
            set2.setCircleColor(Color.WHITE);
            set2.setLineWidth(2f);
            set2.setMode(LineDataSet.Mode.HORIZONTAL_BEZIER);
            set2.setCircleRadius(3f);
            set2.setFillAlpha(65);
            set2.setFillColor(Color.RED);
            set2.setDrawCircleHole(false);
            set2.setHighLightColor(Color.rgb(244, 117, 117));
```

RobertZagorski:
Probably does not happen anymore. I can't reproduce it.

almic:
[...]

PR #4961, opened 2020-08-02
Fix bug #2071, alpha issue of DataSets

qinidea:
## PR Checklist:
       

## PR Description

Issue #2071 has marked closed at 2018 but actually it still open.

LineChartRenderer calls c.drawBitmap with the mRenderPaint , which still contains the alpha from the last drawn dataset.


hannesa2:
I guess, I've already included it  https://github.com/AppDevNext/AndroidChart/blob/268c1b4034e5d397afae60f778930ab92bf30c62/chartLib/src/main/kotlin/info/appdev/charting/renderer/LineChartRenderer.kt#L85

PR #5040, opened 2020-11-09
Update Utils.java

Sublimis:
Use double-s instead of float-s to minimize rounding errors that can lead to axis labels saying e.g. "9999.999" instead of "10000". There's really no excuse for using float-s here, and probably throughout much of the rest of the code, too. Maybe this remark could be taken into consideration to prevent similar problems in the future.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/utils/Utils.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/utils/Utils.java
@@ -352,15 +352,15 @@ public abstract class Utils {
      * @param number
      * @return
      */
-    public static float roundToNextSignificant(double number) {
-        if (Double.isInfinite(number) || 
-            Double.isNaN(number) || 
+    public static double roundToNextSignificant(double number) {
+        if (Double.isInfinite(number) ||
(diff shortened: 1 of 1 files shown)


Paget96:

Code is full kotlin now!

Make your pull requests here:

hannesa2:
@Sublimis
I did a test of this https://github.com/AppDevNext/AndroidChart/pull/637 and there you'll see what happens as screenshots.
Do you have maybe better show cases ?

Sublimis:
@hannesa2
Only found this one old screenshot that shows incorrect rounding on the horizontal axis (1.99k, 3.99k, etc), before this PR fix was applied. Maybe this helps.

image

hannesa2:
@Sublimis 
I [tested it](https://github.com/AppDevNext/AndroidChart/pull/643) in my recent [version 4.0](https://github.com/AppDevNext/AndroidChart/releases/tag/4.0) it looks fine in my point of view https://github.com/AppDevNext/AndroidChart/pull/643/files#diff-b24e02fdba97f1f3339bf857ee57d453e5a7c803ef28a042fe8cce1501928cc5

image

No patch is needed there

PR #5045, opened 2020-11-17
Fixes #405 - added isFlingEnabled setting

### LINKED ISSUE #405: Fling Gesture Not Recognized if Max Visible X range is set.
Hello.

I'm trying to implement an update based on a Fling gesture, where new values are added to the graph as the user swipes left and right. However, onFling is not called back when a max visible X-range is set. This is my current setup:

 mLineChart.setBackgroundColor(getResources().getColor(android.R.color.white));
        mLineChart.setDragEnabled(true);
        mLineChart.setTouchEnabled(true);
        mLineChart.setOnChartGestureListener(mOnChartGestureListener);
        mLineChart.setOnChartValueSelectedListener(mOnChartValueSelectedListener);

.........
.........
 mLineChart.setData(lineData);
        mLineChart.setVisibleXRange(7);

Is there another callback that we should use when performing a fling gesture with a Maximum visible X val set? Thanks.

lesniakania:
## PR Checklist:
- [x] I have added/updated examples and tests for any new behavior (there are no tests/examples for similar things e.g. `isDragEnabled` so I didn't add mine too).

## PR Description

This adds `isFlingEnabled` setting so `onChartFling` can work properly.
Fixes https://github.com/PhilJay/MPAndroidChart/issues/405

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -81,6 +81,11 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
     private boolean mDragXEnabled = true;
     private boolean mDragYEnabled = true;
 
+    /**
+     * if true, fling gesture is enabled for the chart
+     */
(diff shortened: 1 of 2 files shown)


Paget96:

Code is full kotlin now!

Make your pull requests here:

Issue #5054, opened 2020-11-26
BarChart crashes if bar has no values


Athameon:
Situation:
I've got a stacked bar chart where each stack has a different amount of entries.
To get the correct colors (All first entries have the same color, all second entries, ...) I created for each x value bar entry an array that is greater than its values eg: new BarEntry(index, [4, 6, 1, 0, 0, 0, 0, 0, 0, 0, 0, ...].
The size of the array is the size of the available colors, so each BarEntry y value has the same color.
But if I now click into the empty area where no BarEntry exists, it tries to get an invalid BarChart value.

The problem is in BarHighlighter:getStackedHighlight Line 66: entry.getRanges(); which returns null.

CallStack:
E/InputEventReceiver: Exception dispatching input event.
E/MessageQueue-JNI: Exception in MessageQueue callback: handleReceiveCallback
E/MessageQueue-JNI: java.lang.NullPointerException: Attempt to get length of null array
        at com.github.mikephil.charting.highlight.BarHighlighter.getStackedHighlight(BarHighlighter.java:68)
        at com.github.mikephil.charting.highlight.BarHighlighter.getHighlight(BarHighlighter.java:34)
        at com.github.mikephil.charting.charts.BarChart.getHighlightByTouchPoint(BarChart.java:96)
        at com.github.mikephil.charting.charts.BarLineChartBase.getDataSetByTouchPoint(BarLineChartBase.java:1371)
        at com.github.mikephil.charting.listener.BarLineChartTouchListener.saveTouchStart(BarLineChartTouchListener.java:307)
        at com.github.mikephil.charting.listener.BarLineChartTouchListener.onTouch(BarLineChartTouchListener.java:127)
        at com.github.mikephil.charting.charts.BarLineChartBase.onTouchEvent(BarLineChartBase.java:581)
        at android.view.View.dispatchTouchEvent(View.java:9943)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:2663)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2344)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:2669)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2301)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:2669)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2301)
[...]

PR #5055, opened 2020-11-26
Test for null in entry.getRanges()

Athameon:
If a BarChart has no printed values, the method entry.getRanges() returns null and the method ranges.length throws an exception (null.length).
This can simply be prevented by asking the ranges to be not null before asking for length.

## PR Checklist:
- [X] If this is a significant change, an issue has already been created where the problem / solution was discussed: [https://github.com/PhilJay/MPAndroidChart/issues/5054]

## PR Description
Check array to be not null before accessing a property of null. 

The BarChart does not crash if it has empty values.

It prevents a crash of the app.


Paget96:

Code is full kotlin now!

Make your pull requests here:

PR #5074, opened 2021-01-11
Split inner web alpha from all web alpha in RadarChart

itmir913:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N]
       

## PR Description

In RadarChart mWebAlpha, I want to apply the alpha value each.
When draw the web lines and the inner-web, I want to Apply a different value of mWebAlpha.

So, I added the mWebAlphaInner, getWebAlphaInner(), setWebAlphaInner().

for people who want to apply the alpha values each other, I hope it will be merged.

Thanks.


Paget96:

Code is full kotlin now!

Make your pull requests here:

PR #5091, opened 2021-02-07
small bug fix

parthwaidya:
should be entries instead of parties on line 125

## PR Checklist:


MPChartExample/src/main/java/com/xxmassdeveloper/mpchartexample/PieChartActivity.java
```diff
@@ -123,7 +123,7 @@ private void setData(int count, float range) {
         // the chart.
         for (int i = 0; i < count ; i++) {
             entries.add(new PieEntry((float) ((Math.random() * range) + range / 5),
-                    parties[i % parties.length],
+                    entries[i % entries.length],
```
MGaetan89, line 126: `parties` is actually the right one to use here. It is defined in `DemoBase`

hannesa2: Sorry, this will not work. There is is no PieEntry() constructor, which will take a PieEntry as second parameter

hannesa2:
This doesn't work. Here is the evidence https://github.com/AppDevNext/AndroidChart/pull/635

PR #5194, opened 2021-09-20
refactor: improve YAxisRenderer.renderAxisLabels

t49tran:
## PR Checklist:
       

## PR Description
Small refactor to YAxisRenderer by replacing the nested if / else with two functions to calculate the label position x and label text alignment.

Hopefully, this is helpful and improve readability a little bit.


PhilJay:
Thank you for this. It no longer applies to 4.0: `YAxisRenderer` was rewritten in Kotlin, including the label rendering this refactors. Closing.

Issue #5217, opened 2021-11-20
Performance issue with DefaultValueFormatter 


ray26:
In the process of using, it is found that when setting data in the chart, the DefaultValueFormatter will be calculated repeatedly and the DecimalFormat will be initialized. If there are a large number of icons on the interface, this will cause serious performance problems.

If the calculated number of digits has not changed, the DecimalFormat will not be reinitialized

PR #5218, opened 2021-11-20
Improve DefaultValueFormatter performance

larryxiao625:
## PR Checklist:
       

## PR Description

Improve DefaultValueFormatter performance

If the calculated number of digits has not changed, the DecimalFormat will not be reinitialized

As in https://github.com/PhilJay/MPAndroidChart/issues/5217 describe,  the DefaultValueFormatter will be calculated repeatedly and the DecimalFormat will be initialized. If there are a large number of icons on the interface, this will cause serious performance problems.


PhilJay:
Thank you for this. It no longer applies to 4.0: `DefaultValueFormatter` was rewritten in Kotlin. If a hot path is still slow, please open an issue with a measurement. Closing.

PR #5231, opened 2022-01-03
Support dynamically circle colors changes on LineChartRenderer

oalmirgabriel:
## PR Checklist:
       

## PR Description

Improve cache from circle image on LineChartRenderer
Changed mImageCache from HashMap to WeakHashMap to avoid OOM
Moved `DataSetImageCache.init` to be inside of `fill` method. The main idea is check if the color have been changed and update circle if it is true.
### Motivation:
I created this PR because I met one use case that I needed to use different colors from dots and changed it dinamically from a "hover" on graph. So I used the method setCircleColors, but when I call drawCircle it doesn’t works because the length of the circle array(`drawCircles:777`) have not been changed. So, in my use case the size keep the same only the color changed.

Selected:
Screen Shot 2022-01-03 at 18 26 36
Unselected:
Screen Shot 2022-01-03 at 18 26 22


PhilJay:
Thank you. 4.0 covers this: the circle cache now notices a change of colour, radius and hole, so circle colours can be changed at runtime. Closing.

PR #5243, opened 2022-01-15
Fix x-axis labels overlapping

### LINKED ISSUE #2980: Barchart x Axis labels are overlapping
Hi , 
As my x axis labels are long , they are overlapping on each other in bar chart 
There is method **setWordWrapEnabled** available for legends, but , same method is not available for XAxis **mChart.getXAxis()** .

setWordWrapEnabled is support in iOS for xAxis . 

Is there any way to fix this in android ?

### LINKED ISSUE #3298: Overlaping xAxis labels
Hello.

For my bar chart, my xAxis label display the beginning and ending time of a given task(`12:00AM - 1:00AM`). However, the label is too long and overlaps with other labels on the axis. 

untitled

I tired to use the `IndexAxisValueFormatter` where I give an Arraylist of String that has String formatted like this `12:00AM - \n1:00AM`, so I can stack the time on top of each other. That idea did not work. 

Is there a way to get the label to be formatted like that?

### LINKED ISSUE #3673: Issue with bar charts labels
binding.barChart.getXAxis().setValueFormatter(new IndexAxisValueFormatter(labels));

Where labels is a Arraylist of strings. The problem is that the labels are overlap each other also they are starting from very bottom of the barchart view while i want them to move a bit up and also to avoid overlapping.

### LINKED ISSUE #4557: How to make xaxis lables limited
linechart x axis as format (29/05/2000 12:48) such that lables in xaxis are collapsing and there are many points in graph i want at the beginning to show only few points and will scroll to get other points.please help??
Thanks 
Screenshot_20190603-154133

### LINKED ISSUE #4640: text getting cut and overlapped on x axis at the end(line chart)
Untitled-1
text getting cut and overlapped on x axis at the end(line chart) 
check screenshot

yaugenka:
Fixes #3298, Fixes #2980, Fixes #4640, Fixes #4557,  Fixes #3673.

For those who are looking for a workaround solution, have posted one into #3298.

The fix required modification of `AxisRender.ComputeAxisValues` just for `XAxis`.
In order to avoid code duplication, the following refactoring was done.
`ComputeAxisValues` was split into `CompleteAxisInterval` (the part which differs for X and Y axis) and `ComputeAxisValues` (the common part). The resulting workflow is `X/YAxisRenderer.ComputeAxis` -> `X/YAxisRenderer.CompleteAxisInterval` -> `AxisRender.ComputeAxisValues`.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/components/XAxis.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/components/XAxis.java
@@ -48,6 +48,11 @@ public class XAxis extends AxisBase {
     private boolean mAvoidFirstLastClipping = false;
 
     /**
+     * minimum spacing between labels in pixels
+     */
+    protected float mSpaceBetweenLabelsMin = 0;
+
+    /**
      * the position of the x-labels relative to the chart
      */
     private XAxisPosition mPosition = XAxisPosition.TOP;
@@ -63,6 +68,19 @@ public class XAxis extends AxisBase {
         super();
 
         mYOffset = Utils.convertDpToPixel(4.f); // -3
+    }
+
+    /**
+     * Returns minimum spacing between labels in pixels
+     */
+    public float getSpaceBetweenLabelsMin() { return mSpaceBetweenLabelsMin; }
+
+    /**
+     * Sets minimum spacing between labels in pixels
+     */
+    public void setSpaceBetweenLabelsMin(float space)
+    {
+        mSpaceBetweenLabelsMin = space;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


hannesa2:
Nice, what do you think about this ? https://github.com/PhilJay/MPAndroidChart/issues/5268

PhilJay:
Thank you. This is in **4.0.0-beta01**: the x axis keeps its labels at least one label width apart, through a `minimumInterval` hook on `AxisRenderer`. 4.0 is a rewrite in Kotlin, so the Java sources this patch touches no longer exist and it cannot be merged as it stands. It was reimplemented rather than merged, so the credit is yours.

PR #5251, opened 2022-01-26
Fix offsets not adjusted on autoscale

### LINKED ISSUE #2826: Inconsistency of AxisMax and AxisMin
Please see the PR in iOS version 

When enabled autoScale.
data.calcMinMaxY(fromX: self.lowestVisibleX, toX: self.highestVisibleX) is called after
_leftYAxisRenderer?.computeAxis(min: _leftAxis._axisMinimum, max: _leftAxis._axisMaximum, inverted: _leftAxis.isInverted)
_rightYAxisRenderer?.computeAxis(min: _rightAxis._axisMinimum, max: _rightAxis._axisMaximum, inverted: _rightAxis.isInverted)
_xAxisRenderer?.computeAxis(min: _xAxis._axisMinimum, max: _xAxis._axisMaximum, inverted: false)

that means _leftAxisRenderer and _rightAxisRenderer will render AxisLines and Values based on old values of _axisMinimum, _axisMaximum

https://github.com/danielgindi/Charts/pull/2177

### LINKED ISSUE #5237: Chart view is screwed up when recreated
I have a line chart inside main activity fragment layout. When the chart is loaded I zoom it to the last saved scale level.
`zoom(savedScaleFactor, 1f, highestVisibleX, 0f, yAxisDependency)`
The `highestVisibleX` makes it also scrolled to the very right side.

When the application starts and the chart is loaded for the first time it looks good. But when I siwtch to another main activity fragment or start another activity on top and then come back to the charts fragment the yAxis gets screwed up a bit (is shifted out of the screen edge).

mpaissue

If I tap on the chart it gets adjusted to normat state.

mpanoissue

Have tried calling `notifyDataSetChanged()`, `invalidate()`, `moveViewToX(xChartMax)` but nothing helps until a manual tap is performed on the chart. Tried doing `chart.performClick()` programically, but not click passes through. 

Is it possible to perform a click programically somehow?

yaugenka:
Fixes #5237, Fixes #2826.

`calculateOffsets` needs to be called after `computeAxis` because the size of labels can change after the latter.
Also the view port handler needs to be refreshed after that to take the new offsets into account.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/charts/BarLineChartBase.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/charts/BarLineChartBase.java
@@ -210,6 +210,12 @@ public abstract class BarLineChartBase<T extends BarLineScatterCandleBubbleData<
         if (mXAxis.isEnabled())
             mXAxisRenderer.computeAxis(mXAxis.mAxisMinimum, mXAxis.mAxisMaximum, false);
 
+        // Y-axis labels could have changed in size affecting the offsets
+        if (mAutoScaleMinMaxEnabled) {
+            calculateOffsets();
+            mViewPortHandler.refresh(mViewPortHandler.getMatrixTouch(), this, false);
+        }
+
         mXAxisRenderer.renderAxisLine(canvas);
         mAxisRendererLeft.renderAxisLine(canvas);
         mAxisRendererRight.renderAxisLine(canvas);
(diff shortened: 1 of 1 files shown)


Paget96:

Code is full kotlin now!

Make your pull requests here:

PhilJay:
Thank you. This is in **4.0.0-beta01**: the offsets are recomputed after the axis labels, so an auto-scaling chart is no longer a frame behind while panning. 4.0 is a rewrite in Kotlin, so the Java sources this patch touches no longer exist and it cannot be merged as it stands. It was reimplemented rather than merged, so the credit is yours.

PR #5284, opened 2022-04-01
Fix Drawable LineChart Fill Bug

giantramen:
## PR Checklist:
- [x ] I have tested this extensively and it does not break any existing behavior.
       

## PR Description

The LineChart fill draws using intervals to avoid
memory issues, but with a larger data set and many
different (darker) colors Android does not draw the
drawable all the way to the edge of the interval.
This makes it look like there are spaces between the fill.
Draw a little beyond the existing interval so there are no gaps.

Here is what it looks like before my change, look closely at the white vertical lines at ~1/5 ~3/5 ~4/5 of the way across the X-axis.
fill issue

The same color with the fix no longer has gaps.
fill fixed

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/components/AxisBase.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/components/AxisBase.java
@@ -51,6 +51,11 @@ public abstract class AxisBase extends ComponentBase {
      * the number of decimal digits to use
      */
     public int mDecimals;
+    
+    /**
+     * if the user set a custom number of decimals
+     */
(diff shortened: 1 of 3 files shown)


hannesa2:
It's included here https://github.com/AppDevNext/MPAndroidChart/releases/tag/3.1.0.7

PhilJay:
Thank you. This is in **4.0.0-beta01**: the fill chunks overlap by one index when a data set uses a fill drawable, so the seams are gone. 4.0 is a rewrite in Kotlin, so the Java sources this patch touches no longer exist and it cannot be merged as it stands. It was reimplemented rather than merged, so the credit is yours.

Issue #5370, opened 2023-01-03
Scatter shape size treated as pixels, not dp


tgvoskuilen:
**Summary**
The value passed to `setScatterShapeSize` on ScatterDataSet is not converted from dp to pixels in the shape renderers (e.g. CircleShapeRenderer) so the value is treated as a size in pixels, not dp. The scatter shape hole size, however, is correctly converted from dp to pixels. This is present in all the scatter shape renderers.

**Expected Behavior**
All sizes are in dp (as indicated in comments and documentation).

**Possible Solution**
Convert value from dp to pixels in the scatter shape renderers. e.g. change

    final float shapeSize = dataSet.getScatterShapeSize();

to

    final float shapeSize = Utils.convertDpToPixel(dataSet.getScatterShapeSize());

in all the renderers.

**Device (please complete the following information):**
 - Library Version 3.1.0

**Additional Context**
I will submit a PR for this

PhilJay:
This is fixed in **4.0.0-beta01**: The hole radius was converted from dp a second time, so it did not match the shape size. Size and hole share one unit again.

4.0 is a full rewrite in Kotlin. Install instructions and the full list of changes are in the release notes: https://github.com/PhilJay/MPAndroidChart/releases/tag/v4.0.0-beta01

PR #5371, opened 2023-01-03
Fix scatter shape size bug

tgvoskuilen:
Scatter shape sizes were not converted from dp to pixels in the shape renderers so they were drawn at inconsistent size on different density screens.

## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: [[issue link](https://github.com/PhilJay/MPAndroidChart/issues/5370)]
       

## PR Description

Scatter shape size was treated as pixels instead of dp. This is a bug and leads to inconsistent appearance on different density screens. This is also inconsistent with other places (e.g. scatter hole size or line chart marker size) where values were properly treated as dp.

This change fixes the scatter shape renderers so they scale the user-specified size from dp to pixels before drawing the marker.

Using pixels in just one place is a bug. This may cause existing chart markers to change size - and users who were pre-scaling this one dimension by density manually will have to remove that.

### DIFF
--- a/MPChartExample/src/main/java/com/xxmassdeveloper/mpchartexample/custom/CustomScatterShapeRenderer.java
+++ b/MPChartExample/src/main/java/com/xxmassdeveloper/mpchartexample/custom/CustomScatterShapeRenderer.java
@@ -5,6 +5,7 @@ import android.graphics.Paint;
 
 import com.github.mikephil.charting.interfaces.datasets.IScatterDataSet;
 import com.github.mikephil.charting.renderer.scatter.IShapeRenderer;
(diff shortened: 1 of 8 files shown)


hannesa2:
It's included in https://github.com/AppDevNext/AndroidChart/releases/tag/3.1.0.15

PhilJay:
Thank you. This is in **4.0.0-beta01**: the scatter hole radius is no longer converted to pixels twice, so size and hole share one unit. 4.0 is a rewrite in Kotlin, so the Java sources this patch touches no longer exist and it cannot be merged as it stands. It was reimplemented rather than merged, so the credit is yours.

Issue #4489, opened 2019-04-08
Line data chart jumps during zoom


dGorod:
**Summary**
Line data chart jumps during zoom gesture.
Please, check the provided video:
https://youtu.be/QByOeq8UmRY

I have a line chart with two datasets on X-axis. First for the time, second for data. I tried to get into the library's code to understand is it a bug or I'm doing something wrong. But couldn't sort out inner logic from first glance. So I'm looking for the hint what place in the library to look related to such behavior (as in video).

**Device**
 - Device: emulator, Sony Xperia X
 - Android Version: 8.0
 - Library Version: 3.1.0

DropDrage:
I found way to fix this jump bug, but in very weird way. Unfortunately, I couldn't find source of problem. Some debug show, that it seems like you occasionally stop touch and quickly start touch again, but between this actions graph for some reason jumps on some distance on x or y points.

This is a fix, that works for me and adapted for your case(In my case it jumps during Y zoom). Hope it helps you.
```Java
boolean isNoTouchAfterEnd;
MPPointD startScalePoint;

public MPPointD getCenterOfViewPort(){
        final ViewPortHandler handler = chart.getViewPortHandler();
        final float middleX = (handler.contentRight() - handler.contentLeft()) / 2 + handler.contentLeft();
        final float middleY = (handler.contentTop() - handler.contentBottom()) / 2 + handler.contentBottom();

        return chart.getValuesByTouchPoint(middleX, middleY, chart.getAxisLeft().getAxisDependency());
}

    chart.setOnChartGestureListener(new OnChartGestureListener() {
            @Override
            public void onChartGestureStart(MotionEvent me, ChartTouchListener.ChartGesture lastPerformedGesture) {
                startScalePoint = getCenterOfViewPort();
                isNoTouchAfterEnd = false;
            }

            @Override
            public void onChartGestureEnd(MotionEvent me, ChartTouchListener.ChartGesture lastPerformedGesture) {
                if (lastPerformedGesture.equals(ChartTouchListener.ChartGesture.X_ZOOM)) {
                    isNoTouchAfterEnd = true;

                    new Handler().postDelayed(new Runnable() {
                        @Override
                        public void run() {
                            if (isNoTouchAfterEnd)
                                chart.centerViewTo((float) startScalePoint.x,
                                        (float) startScalePoint.y, 
                                        chart.getAxisLeft().getAxisDependency());
                        }
                    }, 50);
                }
            }

            @Override
[...]

PR #5391, opened 2023-05-19
fix after call setMinMaxScaleX, when scaleX to bound value, chart wil…

renzhenfei:
After call setMinMaxScaleX, when operate chart zoom, while scaleX or scaleY scale to max value or min value, the chart will jump.

## PR Checklist:
       #4489

## PR Description
In default BarLineChartTouchListener file, when handle zoom event, source code just post scaleX or scaleY, no check after call postScale the scaleX or scaleY value is valid.

This change limit scale to minScale or maxScale when zoom.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/listener/BarLineChartTouchListener.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/listener/BarLineChartTouchListener.java
@@ -74,6 +74,10 @@ public class BarLineChartTouchListener extends ChartTouchListener<BarLineChartBa
      */
     private float mMinScalePointerDistance;
 
+
+    private final float[] matrixBuffer = new float[9];
+    private final Matrix tempMatrix = new Matrix();
+
     /**
      * Constructor with initialization parameters.
      *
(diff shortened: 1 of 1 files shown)


hannesa2:
Please can you show a before-this-change and a after-this-change screenshot ?

hannesa2:
It will probably merged here https://github.com/AppDevNext/AndroidChart/pull/138

renzhenfei:
> 
before

https://github.com/PhilJay/MPAndroidChart/assets/15669071/faa668d0-de66-4b5e-bf74-cd5d3af8fccb

after


https://github.com/PhilJay/MPAndroidChart/assets/15669071/501f8c71-5bee-4a37-8bc9-4d5ced302d91

hannesa2:
@renzhenfei Thank you I'll merge it

PhilJay:
Thank you. This is in **4.0.0-beta01**: the pinch scale is limited before it is applied, so the chart no longer jumps when it reaches its scale limit. 4.0 is a rewrite in Kotlin, so the Java sources this patch touches no longer exist and it cannot be merged as it stands. It was reimplemented rather than merged, so the credit is yours.

PR #5397, opened 2023-06-14
Fix granularity when force label enabled.

customappdeveloper:
Fix granularity when force label enabled.
Granularity not working when we force enable labels.

## PR Checklist:
- ✓ I have tested this extensively and it does not break any existing behavior.


hannesa2:
It will be merged in https://github.com/AppDevNext/AndroidChart/pull/137

PhilJay:
Thank you. This is in **4.0.0-beta01**: forced labels now stay on the axis granularity and the label count shrinks instead. 4.0 is a rewrite in Kotlin, so the Java sources this patch touches no longer exist and it cannot be merged as it stands. It was reimplemented rather than merged, so the credit is yours.

PR #5413, opened 2023-08-25
Dokumentation Link change so it links to the dokumentation 

lukasb1b:
## PR Checklist:


PhilJay:
Thank you for this. It no longer applies to 4.0: the old documentation site is gone; the guides are at https://philjay.cc/mpandroidchart/docs/. Closing.

PR #5420, opened 2023-10-04
Added RoundedBarChartRenderer & RoundedHorizontalBarChartRenderer for Rounded corner Bar Chart

katara-mohini:
… corner radius positive/negative Bar Chart

## PR Checklist:
- [✓ ] I have tested this extensively and it does not break any existing behavior.
- [✓ ] I have added/updated examples and tests for any new behavior.
- [✓ ] If this is a significant change, an issue has already been created where the problem / solution was discussed: N/A

## PR Description:
1. Added RoundedBarChartRenderer for vertical Positive/Negative BarChart with customised rounded corners. 
2. RoundedHorizontalBarChartRenderer for horizontal BarChart with customised rounded corners. 

## Added Feature:
This PR is for BarChartRenderer and HorizontalBarChartRenderer customisation. 

## PR has to be merged:
So that users can customise BarChart corners according to their requirements.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/RoundedBarChartRenderer.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/renderer/RoundedBarChartRenderer.java
@@ -1,0 +1,324 @@
+package com.github.mikephil.charting.renderer;
+
+import android.graphics.Canvas;
+import android.graphics.LinearGradient;
+import android.graphics.Path;
+import android.graphics.RectF;
+import android.graphics.Shader;
+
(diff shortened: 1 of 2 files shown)


Paget96:
There is crash when we use FloatArray for Y values.

Crash happens here, as we don't have y, instead yVals
```
  if ((dataSet.getEntryForIndex(j / 4).y < 0 && roundedNegativeDataSetRadius > 0)) {
```

This is only needed to make it perfect

PhilJay:
Thank you. 4.0 covers this: rounded bars are built in as `BarDataSet.barCornerRadius`, and `isStackSectionsRounded` rounds every section of a stack. Closing.

PR #5456, opened 2024-05-12
Make MPAndroidChart compatible with Android Studio Iguana (2024)

UrielFrankel:
## PR Checklist:
- [v] I have tested this extensively and it does not break any existing behavior.
- [v] I have added/updated examples and tests for any new behavior.
- [v] If this is a significant change, an issue has already been created where the problem / solution was discussed: [N/A, or add link to issue here]


hannesa2:
Btw, it's already included here https://github.com/AppDevNext/AndroidChart/pull/208
This is a maintained fork. It comes with screenshot compare tests as well

PR #4502, opened 2019-04-20
fix Poolable object leak context issue

lbtrace:
## PR Description

- As we know ObjectPool cache some Poolable Objects, but some Poolable Object reference to View 
Object, and View Object reference to Context Object, And ObjectPool is static, so leak Context. For exsmple, ObjectPool.objects ---> AnimatedMoveViewJob.view ---> View.mContext ---> Context

- For this issue, I Clear Poolable Object fields when Poolable Object is recycled. Please check patch code

- I think lots of memory leak occurs when using MPAndroidChart, so please merge this patch.

PR #5511, opened 2025-07-02
Fix the problem that the pool static property holds the View referenc…

qiduhe:
…e, which causes Activity memory leak

## PR Checklist:


hannesa2:
duplicate of https://github.com/PhilJay/MPAndroidChart/pull/4502

mohmmadbd90-cmyk (approved):

PhilJay:
Thank you. 4.0 covers this: the leak behind this is real and is fixed in 4.0, differently: the object pools have an upper bound and the viewport jobs are capped. Closing.

Issue #5516, opened 2025-12-03
Native Gantt Chart / Horizontal Time Bar Support in MPAndroidChart


Dylan-lijl:
**Is your feature request related to a problem? Please describe.**
MPAndroidChart currently supports bar charts, line charts, pie charts, etc., but it lacks native support for **Gantt charts** or horizontal bars representing time intervals.  

We often need charts where:
- Y-axis represents tasks or steps
- X-axis represents time
- Each task/step is visualized as a horizontal rectangle spanning from start time to end time
- Optionally, stacked segments, colors, labels, and values can be shown

Currently, achieving this requires writing a custom renderer, which is cumbersome and error-prone.

**Describe the solution you'd like**
Native support for a **Gantt chart** or **horizontal time bar chart** in MPAndroidChart that:
- Supports horizontal bars representing start/end times
- Handles labels, stacked segments, colors, and value display
- Allows easy configuration without a custom renderer

**Describe alternatives you've considered**
We have implemented a custom renderer extending `HorizontalBarChartRenderer` to simulate this feature. Example:

```java
public class SingleStackRender extends HorizontalBarChartRenderer {
    private RectF mBarShadowRectBuffer = new RectF();

    public SingleStackRender(BarDataProvider chart, ChartAnimator animator,
                             ViewPortHandler viewPortHandler) {
        super(chart, animator, viewPortHandler);
    }

    @Override
    public void drawDataSet(Canvas c, IBarDataSet dataSet, int index) {
        Transformer trans = mChart.getTransformer(dataSet.getAxisDependency());
        BarBuffer buffer = mBarBuffers[index];
        float phaseX = mAnimator.getPhaseX();
        float phaseY = mAnimator.getPhaseY();

        buffer.setPhases(phaseX, phaseY);
        buffer.setDataSet(index);
        buffer.setInverted(mChart.isInverted(dataSet.getAxisDependency()));
        buffer.setBarWidth(mChart.getBarData().getBarWidth());

        feedSingleStackBuffer(buffer, dataSet);
        trans.pointValuesToPixel(buffer.buffer);

        final boolean isSingleColor = dataSet.getColors().size() == 1;

        for (int j = 0; j < buffer.size(); j += 4) {
            if (!mViewPortHandler.isInBoundsTop(buffer.buffer[j + 3])) break;
            if (!mViewPortHandler.isInBoundsBottom(buffer.buffer[j + 1])) continue;

[...]

PR #5522, opened 2026-04-03
Gantt feature

Dayvid-T:
## PR Checklist:
- [x] If this is a significant change, an issue has already been created where the problem / solution was discussed: #5516 

### What does this add/remove/fix/change?  

Adds native Gantt chart support via a new `GanttChart` view. Includes task model, data container, utils, and basic docs.  

**Key Features:**  
- Time-based horizontal bars  
- Scales with number of tasks  
- Custom colors and labels  
- Simple grid/time axis  

The feature Fills a missing feature in the library. Provides a clean built-in solution instead of workarounds. Doesn't break anything and follows existing patterns.

### DIFF
--- a/MPChartLib/src/main/java/com/github/mikephil/charting/charts/GanttChart.java
+++ b/MPChartLib/src/main/java/com/github/mikephil/charting/charts/GanttChart.java
@@ -75,10 +75,10 @@ public class GanttChart extends View {
     }
 
     private void calculateDimensions() {
-        chartLeft = padding + 100; // Leave space for task labels
-        chartTop = padding + 50;
+        chartLeft = padding + 70; // Smaller left margin
+        chartTop = padding + 30;
         chartRight = getWidth() - padding;
-        chartBottom = getHeight() - padding - 50;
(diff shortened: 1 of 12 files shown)


MPChartLib/src/main/java/com/github/mikephil/charting/charts/GanttChart.java
```diff
@@ -0,0 +1,159 @@
...
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.graphics.RectF;
+import android.util.AttributeSet;
+import android.view.View;
+
+import com.github.mikephil.charting.data.GanttChartData;
+import com.github.mikephil.charting.data.GanttTask;
+
+public class GanttChart extends View {
```
hannesa2, line 13: it doesn't extends `Chart`

hannesa2 (commented):
The labels were not good, but I adapt it
image

PhilJay:
A Gantt chart needs its own data type, renderer and highlighter. Out of scope for this library, same as #5516. Thank you for taking the time. Closing.

PR #5524, opened 2026-06-10
Optimized interaction touch-target mechanics across multiple chart families

venkatbhoogarbh:
Description
This PR optimizes the interaction target engine inside ChartHighlighter and PieRadarHighlighter to resolve several outstanding touch-target and layout friction bugs across multiple chart families:

1. BarChart: Maintained strict column pillar boundaries so that touches are constrained precisely to the drawn geometric vertical bars, eliminating overlapping column ghost-clicks.
2. LineChart & ScatterChart: Replaced rigid raw pixel values with a responsive, device-agnostic density-scaled 40dp finger-sized target bubble around vertices for crisp selections on high-DPI displays.
3. BubbleChart: Implemented a dynamic z-index boundary check. Touches are strictly bound to the true physical drawn radius of each individual bubble, and smaller/nested overlapping entries are prioritized using a 15dp touch target floor to prevent large background elements from hijacking taps.
4. PieChart: Expanded the touch boundary perimeter proportionally (1.25x radius) to ensure floating external data text labels and selection lines remain completely responsive to user inputs.
5. RadarChart: Added a density-scaled 40dp vertex-snapping proximity filter to prevent distant accidental selections in empty space.

All modifications are safely type-guarded explicitly using instanceof checks to eliminate any cross-contamination or regressions across independent graph families.


hannesa2:
Nice, but did you looked on the git history here ? I assume that this will never merged here.
If you really want something merged, then please open a similar pull request [here](https://github.com/AppDevNext/AndroidChart) in this Kotlin-only fork, there it will be merged.

PhilJay:
The idea has merit, but the patch carries debug logging and scratch comments, and tapping anywhere in a bar's column is intended behaviour. Thank you for taking the time. Closing.

